PR #11, opened 2017-02-16
Update doc for ContinuousCallback

crbinz:


docs/src/features/callback_functions.md
```diff
@@ -462,9 +462,7 @@ to be their current length + 1, growing the ODE system. Then the following code
 sets the new protein concentrations. Now we can solve:
 
 ```julia
-rootfind = true
-save_positions = (true,true)
-callback = ContinuousCallback(condition,affect!,rootfind,save_positions)
+callback = ContinuousCallback(condition,affect!,rootfind = true,save_positions = (true,true))
```
ChrisRackauckas: This can just be `callback = ContinuousCallback(condition,affect!)` since it matches the defaults.

ChrisRackauckas:
Thanks!

PR #13, opened 2017-03-04
Fixed typo in parameter_estimation.md

Ayush-iitkgp:
result.minimum[1] returns the minimum value of the objective function whereas result.minimizer[1] returns that value of the variable at which the minimum occurs.


ChrisRackauckas:
Thanks!

Issue #16, opened 2017-03-18
Docs for split coupling


ChrisRackauckas:
This needs to be added to the docs @elevien

elevien:
I'll add something within the next week or so.

PR #17, opened 2017-03-27
documentation for split coupling 

elevien:
In response to #16. Is this ok? I can add more detail, such as a mathematical description of the split coupling if you'd like.


ChrisRackauckas:
Looks good. Thanks! What's the paper for the algorithm you used? I should probably add a citation to https://github.com/JuliaDiffEq/juliadiffeq.github.io/blob/master/citing.md

elevien:
> What's the paper for the algorithm you used?

David F. Anderson, Masanori Koyama; An asymptotic relationship between coupling methods for stochastically modeled population processes. IMA J Numer Anal 2015; 35 (4): 1757-1778. doi: 10.1093/imanum/dru044

PR #26, opened 2017-05-09
Update plot.md

aytekinar:
Some typographical changes.

### DIFF
--- a/docs/src/basics/plot.md
+++ b/docs/src/basics/plot.md
@@ -52,17 +52,19 @@ master form is:
 vars = [(f1,0,1), (f2,1,3), (f3,4,5)]
 ```
 
-where this would mean plot f1(variable 1, variable 2), f2(variable 1, variable 3),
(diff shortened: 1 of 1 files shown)


docs/src/basics/plot.md
```diff
@@ -61,7 +61,7 @@ a tuple. If no function is given, for example,
 vars = [(0,1), (1,3), (4,5)]
 ```
 
-where this would mean plot variable 0 vs 1, 1 vs 3, and 4 vs 5 all on the same graph.
+where this would mean plot variable 1 vs 0, 3 vs 1, and 5 vs 4 all on the same graph.
```
ChrisRackauckas: Maybe there is a less ambiguous way to state this?

aytekinar:
I hope you would like this version. Maybe a bit more verbose than needed?

ChrisRackauckas:
Maybe a little verbose, but it's unambiguous and clear. Merging when the build finishes. Thanks!

PR #41, opened 2017-08-27
Add BVP example

YingboMa:


ChrisRackauckas:
You need to add it to the index and make files too.

docs/src/tutorials/bvp_example.md
```diff
@@ -0,0 +1,76 @@
...
+end
+```
+
+### Boundary Condition
+
+And here is where the `Boundary` comes in. We need to write a function that calculate the residual in-place from the problem solution, such that the residual is $\vec{0}$ when the boundary condition is satisfied.
+
+```julia
+function bc1(residual, u)
+    residual[1] = u[end÷2][1] + pi/2 # the solution at the middle of the time span should be -pi/2
```
ChrisRackauckas: This isn't necessarily the middle with adaptive timestepping in shooting. You should change that one to use the interpolation.

docs/src/tutorials/bvp_example.md
```diff
@@ -64,7 +64,12 @@ We have used the mono-implicit Runge–Kutta (MIRK) method to solve the BVP, but
 ```julia
 using OrdinaryDiffEq
 u₀_2 = [-1.6, -1.7] # the initial guess
-sol3 = solve(bvp1, Shooting(Vern7()))
+function bc3(residual, u)
```
ChrisRackauckas: You should explain here that for `Shooting`, `u` is the "current solution". In fact, you may want to also change the name to `sol` to make that more explicit.

PR #45, opened 2017-09-07
Update ode_solve.md

ranocha:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/187

### DIFF
--- a/docs/src/solvers/ode_solve.md
+++ b/docs/src/solvers/ode_solve.md
@@ -180,7 +180,7 @@ Zhang and Shu (Zhang, Xiangxiong, and Chi-Wang Shu. "Maximum-principle-satisfyin
 high-order schemes for conservation laws: survey and new developments." Proceedings of the Royal Society of
 London A: Mathematical, Physical and Engineering Sciences. The Royal Society, 2011.).
 
-### Low-Storage Methods for Hyperbolic PDEs (Conservation Laws)
+### Low-Storage Methods (Designed for Hyperbolic PDEs, Conservation Laws)
 
 - `CarpenterKennedy2N54` - The five-stage, fourth order low-storage method of Carpenter and Kennedy 
   (free 3rd order Hermite interpolant). Fixed timestep only.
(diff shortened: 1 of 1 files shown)


docs/src/solvers/ode_solve.md
```diff
@@ -180,6 +180,11 @@ Zhang and Shu (Zhang, Xiangxiong, and Chi-Wang Shu. "Maximum-principle-satisfyin
 high-order schemes for conservation laws: survey and new developments." Proceedings of the Royal Society of
 London A: Mathematical, Physical and Engineering Sciences. The Royal Society, 2011.).
 
+### Low-Storage Methods for Hyperbolic PDEs (Conservation Laws)
```
ChrisRackauckas: Are these specifically for Hyperbolic PDEs? I thought low storage methods are just in general for large systems.

ranocha: Yes, they can be used in in general for large system, but they have been designed especially for hyperbolic equations (since the authors are interested in conservation laws).

ChrisRackauckas: What about these are specific to hyperbolic equations?

ChrisRackauckas:
I think we should only mention that they were made for hyperbolic PDEs if there's something about the algorithm(s) which is specifically for hyperbolic PDEs. What feature other than low storage (which is more general) helps them be efficient in this case?

ranocha:
Their stability region has been designed with hyperbolic PDEs in mind.

ChrisRackauckas:
I see. But that's just this one method? I think this can receive a note, but the section header on low storage methods shouldn't mention it, because I believe most other low-storage methods are not hyperbolic PDE specific.

ChrisRackauckas:
How exactly is the stability region hyperbolic PDE specific? Something to do with the complex part?

ranocha:
Yes, balancing the stability region on the imaginary and real axis.

PR #51, opened 2017-09-30
Fix ODE tutorial link

liuhenry:
I was reading the getting started page and noticed this link was broken. Looks like it should be a relative path.


ChrisRackauckas:
Thanks!

PR #68, opened 2018-02-01
Correct

BeastyBlacksmith:
This sentence would imply that by default all steps get rejected.

### DIFF
--- a/docs/src/basics/common_solver_opts.md
+++ b/docs/src/basics/common_solver_opts.md
@@ -176,9 +176,9 @@ explanations of the timestepping algorithms, see the
 * `callback`: Specifies a callback. Defaults to a callback function which
   performs the saving routine. For more information, see the
   [Event Handling and Callback Functions manual page](../../features/callback_functions.html).
-* `isoutofdomain`: Specifies a function `isoutofdomain(t,u)` where, when it
+* `isoutofdomain`: Specifies a function `isoutofdomain(u,t)` where, when it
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
What it means to say is that `isoutofdomain(u,p,t) = false` is the default (this part needs the argument update)

docs/src/basics/common_solver_opts.md
```diff
@@ -176,9 +176,9 @@ explanations of the timestepping algorithms, see the
 * `callback`: Specifies a callback. Defaults to a callback function which
   performs the saving routine. For more information, see the
   [Event Handling and Callback Functions manual page](../../features/callback_functions.html).
-* `isoutofdomain`: Specifies a function `isoutofdomain(t,u)` where, when it
+* `isoutofdomain`: Specifies a function `isoutofdomain(u,t)` where, when it
```
ChrisRackauckas: u,p,t. It's seen here: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/common_defaults.jl

BeastyBlacksmith:
I hope i got it all right now

ChrisRackauckas:
Actually it looks like somewhere along the line the default for that was changed to check for NaNs.

BeastyBlacksmith:
true, i'll change that

BeastyBlacksmith:
no, thats `unstable_check` that does it. So everything should be fine now.

ChrisRackauckas:
Thanks!

PR #69, opened 2018-02-01
Update callback_library.md

BeastyBlacksmith:
corrected logical issue.

### DIFF
--- a/docs/src/features/callback_library.md
+++ b/docs/src/features/callback_library.md
@@ -112,7 +112,7 @@ satisfies this property mathematically it can be difficult for ODE solvers to
 ensure it numerically, as these [MATLAB examples](https://www.mathworks.com/help/matlab/math/nonnegative-ode-solution.html)
 show.
 
-In order to deal with this problem one can specify `isoutofdomain=(u,p,t) -> any(x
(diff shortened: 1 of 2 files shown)


docs/src/features/callback_library.md
```diff
@@ -112,7 +112,7 @@ satisfies this property mathematically it can be difficult for ODE solvers to
 ensure it numerically, as these [MATLAB examples](https://www.mathworks.com/help/matlab/math/nonnegative-ode-solution.html)
 show.
 
-In order to deal with this problem one can specify `isoutofdomain=(u,p,t) -> any(x
+In order to deal with this problem one can specify `isoutofdomain=(u,p,t) -> !any(x
```
ChrisRackauckas: if it's ever true, it will reject. The original way of writing will have it reject when any term is negative.

BeastyBlacksmith: From the docs:
> isoutofdomain: Specifies a function isoutofdomain(t,u) where, when it returns false, it will reject the timestep. Defaults to always false.

One of them is wrong.

ChrisRackauckas: Oh, that should say "when it returns true"

BeastyBlacksmith:
ok, corrected that.

PR #71, opened 2018-02-09
Update ode_example.md

onoderat:
Found another small instance where the breaking change convention in specifying the problem instance needs to be updated in the docs.


ChrisRackauckas:
Thanks!

PR #79, opened 2018-02-15
Added steps to access plotting and summary statistics

Vaibhavdixit02:


docs/src/analysis/parameter_estimation.md
```diff
@@ -906,4 +917,23 @@ The chain for the `i`th parameter is given by:
...
+```julia
+DynamicHMC.NUTS_statistics(bayesian_result_dynamic[2])
+```
+Some details about the NUTS sampler can be obtained from:
+
+```julia
+bayesian_result_dynamic[3]
+```
+
+Incase of `dynamic_inference` the trace plots for the `i`th parameter can be obtained by:
```
ChrisRackauckas: In case

PR #89, opened 2018-02-21
Document remake

### LINKED ISSUE #84: Document problem remake function
@tkf where should it go?

tkf:
closes #84 
blocked by https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/92

Not sure the second commit about SplitFunction is worth mentioning it, though. Does anybody want to change SplitFunction?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -24,7 +24,7 @@ makedocs(modules=[DiffEqPDEBase,DiffEqProblemLibrary],
            "basics/solution.md",
            "basics/plot.md",
            "basics/integrator.md",
-           "basics/structs.md",
+           "basics/problem.md",
            "basics/faq.md",
            "basics/compatibility_chart.md"
          ],
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -24,6 +24,7 @@ makedocs(modules=[DiffEqPDEBase,DiffEqProblemLibrary],
            "basics/solution.md",
            "basics/plot.md",
            "basics/integrator.md",
+           "basics/structs.md",
            "basics/faq.md",
            "basics/compatibility_chart.md"
          ],
--- a/docs/src/basics/structs.md
+++ b/docs/src/basics/structs.md
@@ -1,0 +1,26 @@
+# Immutable `struct` Handling
+
+Problem-related types in DifferentialEquations.jl are immutable.  This
+helps, e.g., parallel solvers to efficiently handle problem types.
+
+However, you may want to modify the problem after it is created.  For
+example, to simulate it for longer timespan.  It can be done by the
+`remake` function:
+
+```julia
+prob1 = ODEProblem((u,p,t) -> u/2, 1.0, (0.0,1.0))
+prob2 = remake(prob1; tspan=(0.0,2.0))
+```
+
+A general syntax of `remake` is
+
+```julia
+modified_struct = remake(original_struct;
+  field_1 = value_1,
+  field_2 = value_2,
+  ...
+)
+```
+
+where `field_N` and `value_N` are renamed to appropriate field names
+and new desired values.
--- a/docs/src/basics/problem.md
+++ b/docs/src/basics/problem.md
@@ -1,4 +1,6 @@
-# Immutable `struct` Handling
+# Problem interface
+
+## Modification of problem types
 
 Problem-related types in DifferentialEquations.jl are immutable.  This
 helps, e.g., parallel solvers to efficiently handle problem types.
@@ -15,7 +17,7 @@ prob2 = remake(prob1; tspan=(0.0,2.0))
 A general syntax of `remake` is
 
 ```julia
-modified_struct = remake(original_struct;
+modified_problem = remake(original_problem;
   field_1 = value_1,
   field_2 = value_2,
   ...
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -131,6 +131,7 @@ Pages = [
     "basics/solution.md",
     "basics/plot.md",
     "basics/integrator.md",
+    "basics/problem.md",
     "basics/faq.md",
     "basics/compatibility_chart.md"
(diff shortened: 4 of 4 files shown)


docs/src/basics/structs.md
```diff
@@ -0,0 +1,37 @@
+# Immutable `struct` Handling
```
ChrisRackauckas: Instead of being about structs, this page should be specifically about handling the problem types.

docs/src/basics/structs.md
```diff
@@ -0,0 +1,37 @@
...
+
+Problem-related types in DifferentialEquations.jl are immutable.  This
```
ChrisRackauckas: Subsection on remake

ChrisRackauckas:
We should flesh this out a bit. What things should people generally know about the problem types?

1. Type specification. `uType`, `tType`, and things like prototype types (noise_rate_prototype, jac_prototype, etc.) and parameters all come from the passed values. These can be any type. Example of a struct for parameters?
2. Type-equivalencies (SplitODEProblem -> ODEProblem with a SplitFunction)? Or is this an implementation detail?
3. `isinplace` and setting `inplace`.

?

docs/src/basics/structs.md
```diff
@@ -0,0 +1,37 @@
...
+  field_2 = value_2,
+  ...
+)
+```
+
+where `field_N` and `value_N` are renamed to appropriate field names
+and new desired values.
+
+Note that `remake` is not restricted to problem types.  For example,
+`SplitFunction` used via `SplitODEProblem` can be modified by `remake`
```
ChrisRackauckas: `SplitFunction` isn't documented anywhere else. I am wondering if I should just be considering it an internal implementation detail.

tkf: I think it's better to expose less when it comes to API.  So let's remove it from the first version of the document.

docs/src/basics/structs.md
tkf: So rename structs.md to problem.md?

docs/src/basics/structs.md
tkf: You mean, something like:

> # Problem types
> 
> Problem-related types in DifferentialEquations.jl are immutable.  This
> helps, e.g., parallel solvers to efficiently handle problem types.
> 
> ## `remake``

?

docs/src/basics/structs.md
ChrisRackauckas: yeah and change the title.

ChrisRackauckas:
We should probably address functional initial conditions/timespans here too.

tkf:
> Type-equivalencies (SplitODEProblem -> ODEProblem with a SplitFunction)? Or is this an implementation detail?

To be very honest, `SplitODEProblem(...)` returning an `ODEProblem` is a bit of WTF moment when I was reading the code :wink:.  I can imagine that it probably just was a hacky implementation to make the downstream packages work transparently, or something like that.  But I expect `SomeType(...) isa SomeType` to hold in general.

Putting my impression aside, it is possible to make `remake` for `SplitODEProblem` "flat", i.e, calling `remake(split_problem; f1=..., f2=...)` work, even with current type/struct hierarchy.  The idea is to allow "interception" of constructor lookup by:

```julia
function constructor_of(thing)
  T = parameterless_type(thing)
  return if method_exists(T, ())
    # This path is required for, e.g., NoiseProblem
    T
  else
    # Assume that T wants isinplace
    T{isinplace(thing)}
  end
end

constructor_of(prob::ODEProblem{...type parameters..., <:SplitODEProblem}) =
    SplitODEProblem

function remake(thing; kwargs...)
  constructor = constructor_of(thing)
  return constructor(; struct_as_dict(thing)..., kwargs...)
end
```

docs/src/basics/structs.md
ChrisRackauckas: Yeah.

tkf:
> functional initial conditions/timespans

OK. So I'll put `remake` stuff entirely in a sub-section so that it'll be easy to add them.

tkf:
Actually, I don't know whether or not `SomeType(...) isa SomeType` is the invariance that Julians typically expect.  It was coming from my experience in other languages.

ChrisRackauckas:
A constructor is just a function. But yes, it is a little bit of abuse. It does set a unique trait though. It is so that way it can be used downstream in "standard ODE solvers" without a hiccup though.

tkf:
> Type specification. uType, tType, and things like prototype types (noise_rate_prototype, jac_prototype, etc.) and parameters all come from the passed values. These can be any type. Example of a struct for parameters?

Aren't they specific to each problem type?  Or you mean you want document a minimal interface that all problem types satisfy?

> isinplace and setting inplace.

Does it mean to move http://devdocs.juliadiffeq.org/latest/contributing/type_traits.html to here (problem.md)?

tkf:
I did "speculative executions/edits" assuming the answer to my questions above are all yes :)

docs/make.jl
```diff
@@ -24,6 +24,7 @@ makedocs(modules=[DiffEqPDEBase,DiffEqProblemLibrary],
            "basics/solution.md",
            "basics/plot.md",
            "basics/integrator.md",
+           "basics/structs.md",
```
ChrisRackauckas: This needs a change. Also it should get a homepage link

tkf: Oops. Thanks!

tkf: What do you mean by homepage link?

ChrisRackauckas: The index of the docs summarizes each section and lists the pages.

tkf: Like this? https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/89/commits/8703915d9de64128fdd9909b35e1a31cb2b51e35

PR #105, opened 2018-03-14
Documentation tweak

JackDevine:
The in-place functions write into the first variable. Also, the analytical solution does not write into anything.


ChrisRackauckas:
Thanks

PR #106, opened 2018-03-16
fix typo on BVP example page

MaximilianJHuber:
In `bc2!` the variable `u[1]` is a two-vector, not a scalar. Running the code as is yields a MethodError.

I select the first element of the two-vector in each boundary condition and get the same plot as on the page.


ChrisRackauckas:
Thanks!

PR #112, opened 2018-03-23
corrected definitions of the objective functions

Mikhail-Vaganov:
Brought the documentation in accordance with tests and actual interfaces.
Also corrected the remark upon the quadratic form of the `du/dt` function.


ChrisRackauckas:
Thanks

PR #113, opened 2018-03-24
Update DAE tutorial to show bump in y2

non-Jedi:
Something funky goes on if you just use the
plot(sol, xlims=(1e-6,1e6), xscale=:log10, layout=(3,1)) form. Nothing
gets plotted before about 100 on the x-axis. This works around that.


ChrisRackauckas:
The problem is that xlims is ignored for tspan for the continuous sampling. Maybe they should alias but tspan in the recipe should be fine

non-Jedi:
Hm. Okay. I dug in a little bit more. Unless you set the `plotdensity` super
frickin' high, you're always gonna have bad looking plots when using
`denseplot`. This is because the plotting recipe always calls `linspace` on the
tspan. I'm gonna put this PR on hold until we can get something into the
plotting recipe for using `logspace` when necessitated by `xscale`.

ChrisRackauckas:
>This is because the plotting recipe always calls linspace on the tspan.
 
Oh... I did not consider this. Yes, it's an issue.

non-Jedi:
@ChrisRackauckas this is updated and ready now.

ChrisRackauckas:
Thanks!

PR #115, opened 2018-03-25
fixed a typo

Mikhail-Vaganov:
integrator is an iterator


ChrisRackauckas:
hehe, it was true before though 👍

PR #116, opened 2018-03-25
Update history function in DDE docs

devmotion:
This PR updates history functions in the DDE documentation to the new interface with parameters.


ChrisRackauckas:
Thanks! This is a very good interface update. Thanks for the contribution.

PR #117, opened 2018-03-26
Fix a typo in Ornstein-Uhlenbeck process equation

tkf:


ChrisRackauckas:
The first one confused me, but yes thanks for noticing.

tkf:
Doing everything in browser was hard...

PR #121, opened 2018-04-06
Typo in performance_overloads

tobydriscoll:
The documentation for the DAE Jacobian uses `gamma` differently than the provided example. It's the example, and not the stated version, that appears to work for me.


ChrisRackauckas:
Thanks. I flipped it so it looks more like the `inv_W` setup of the ODEs. In a sense, they are the same thing. But it looks like the way DAEs is done is slightly differently, and should change that to match Sundials' DAE setup...

PR #127, opened 2018-04-14
JumpProblem - MassActionJump updates

isaacsas:
1. Added note explaining how to create zeroth order reactions
2. Added note explaining that species indices within stoichiometry vectors should be ordered from smallest to largest index.


ChrisRackauckas:
>Added note explaining that species indices within stoichiometry vectors should be ordered from smallest to largest index.

Why is that true?

isaacsas:
I don't think it would actually break anything, but I was thinking it could have performance implications. I also haven't tested whether passing species out of order could have any issues.

ChrisRackauckas:
Oh yes, it would have performance implications given the indexing order. But the way it's written it sounds like something could break.

isaacsas:
I could change it from "should" be ordered to "are recommended".

isaacsas:
Ok, fixed the wording.

PR #129, opened 2018-05-03
obj is not defined yet here, cost_function works

rafaqz:


ChrisRackauckas:
yes, oops. That was a mistake. Thanks!

PR #133, opened 2018-06-12
emojifiy LotkaVolterra

miguelraz:


oxinabox:
That use of Emoji is actually clearer than normal variable names.
Just saying

Datseris:
I think this should be merged.

PR #135, opened 2018-06-18
Update bvp_example.md

Firemage0520:
I misclicked enter before I could add a comment, whoops. Running the code in the BVP tutorial section for the shooting method example produces a method error, adding 't' to the argument of bc3! fixes it.


ChrisRackauckas:
Thanks!

PR #136, opened 2018-07-06
Update the integrator interface documentation

dawbarton:
It seems like a change from `userdata` to `p` in the integrator interface hasn't been updated in the documentation. Also, while the other fields are described by their values, `userdata` was described by its type which is slightly odd, so I've adjusted it to describe it by its value.


ChrisRackauckas:
Thanks!

PR #138, opened 2018-07-16
Update biological.md

summer-mute:
added rType to MM and birth death examples


ChrisRackauckas:
@Gaussia is the type name required or is it gensym'd?

ChrisRackauckas:
Thanks for the clarifications!

PR #139, opened 2018-07-17
Explicitly naming DEIntegrator in docs

artkuo:
The documentation doesn't explicitly name the composite type `DEIntegrator`. It refers mainly to the instance `integrator` (and sometimes `Integrator`). I've therefore added `DEIntegrator` in a few places. (I needed to know because I wanted to set a dispatch pattern for a callback condition.) 

On another note, there is a discrepancy, where the documentation shows the prototype `reinit!(integrator::ODEIntegrator …)`, whereas the code docstring says `reinit!(integrator::DDEIntegrator …`. I suspect these should be made consistent with each other, but I wasn't sure whether it should be `DEIntegrator`/`ODEIntegrator`/`DDEIntegrator`. My guess is `DEIntegrator` since that is most general, and `reinit!` seems to be quite general. But I wasn't sure, so I didn't change it.

### DIFF
--- a/docs/src/basics/integrator.md
+++ b/docs/src/basics/integrator.md
@@ -301,7 +301,7 @@ solve!(integrator)
 
 ## Plot Recipe
 
-Like the `Solution` type, a plot recipe is provided for the `DEIntegrator` type.
+Like the `DESolution` type, a plot recipe is provided for the `DEIntegrator` type.
 Since the `DEIntegrator` type is a local state type on the current interval,
 `plot(integrator)` returns the solution on the current interval. The same
 options for the plot recipe are provided as for `sol`, meaning one can choose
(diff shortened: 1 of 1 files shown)


docs/src/basics/integrator.md
```diff
@@ -14,7 +14,7 @@ integrator = init(prob,alg;kwargs...)
 ```
 
 The keyword args which are accepted are the same [Common Solver Options](@ref)
-used by `solve`. The type which is returned is the integrator. One can manually
+used by `solve`. The type which is returned is an `integrator<:DEIntegrator`. One can manually
```
ChrisRackauckas: The return value is `typeof(integrator) <: DEIntegrator`.

ChrisRackauckas:
I like the idea, but I think we should be careful with `integrator` the value vs `DEIntegrator` the type.

ChrisRackauckas:
Thanks!

PR #140, opened 2018-07-17
More referencing of integrator interface

artkuo:
In the `condition` description, I added a reference to the manual page on integrator interface. This is to make it more explicit, since `integrator` is not described until the next bullet point.


ChrisRackauckas:
Thanks! I like it.

PR #147, opened 2018-08-13
Update Biochemical Reaction Networks DSL Docs

TorkelE:
Made the DSL documentation more extensive and somewhat more orderly.


docs/src/models/biological.md
```diff
@@ -1,166 +1,307 @@
 # Chemical Reaction Models
 
-The biological models functionality is provided by DiffEqBiological.jl and helps
-the user build discrete stochastic and differential equation based systems
+The biological models functionality is provided by `DiffEqBiological.j`l and helps
```
ChrisRackauckas: mistake?

docs/src/models/biological.md
```diff
@@ -1,166 +1,307 @@
...
+the user to build discrete stochastic and differential equation based systems
 biological models. These tools allow one to define the models at a high level
 by specifying reactions and rate constants, and the creation of the actual problems
-is then handled by the modeling package.
+is then handled by the modelling package.
```
ChrisRackauckas: it should be modeling

docs/src/models/biological.md
```diff
@@ -1,166 +1,307 @@
...
-where each line corresponds to a chemical reaction. The input `rType` designates
-the type of this instance (all instances will inherit from the abstract type
-`AbstractReactionNetwork`).
+where each line corresponds to a chemical reaction. Each reaction consists of a reaction rate (the expression on the left hand side of  `,`), a set of substrates (the expression in-between `,` and `-->`), and a set of products (the expression on the right hand side of `-->`). The substrates and the products may contain one or more reactants, separated by `+`.  The naming convention for these are the same as for normal variables in Julia.
 
-The DSL can handle several types of arrows, in both backwards and forward
-direction. If a bi-directional arrow is used two reaction rates must be
-designated. These two reaction networks are identical:
+The chemical reaction model is generated by the `@reaction_network` macro and stored in the `rn` variable (a normal variable, do not need to be called `rn`). The macro generates a differential equation model according to the law of mass action, in the above example the ODEs become:
+d[X]/dt = -2.0\*[X\]*[Y]
```
ChrisRackauckas: Put this in literal form via ```, or use TeX

docs/src/models/biological.md
```diff
@@ -1,166 +1,307 @@
...
+Just as when defining normal differential equations using `DifferentialEquations` parameter values does not need to be set when the model is created. Components can be designated as parameters by declaring them at the end:
+```julia
+rn = @reaction_network begin
+  p, ∅ → X
+  d, X → ∅
+end p d
+```
+Parameters can only exist in the reaction rates (where they can be mixed with reactants). All variables not declared at the end will be considered a reactant.
 
+#### Pre-defined functions
```
ChrisRackauckas: I think you might need a space under the headers for them to parse correctly.

docs/src/models/biological.md
TorkelE: Yes.

docs/src/models/biological.md
TorkelE: You are right.

docs/src/models/biological.md
TorkelE: You are right, that becomes much better. Can one use TeX in these documents?

ChrisRackauckas: yes, you can use TeX. Just make it a ```math environment

docs/src/models/biological.md
TorkelE: Unsure, I checked everything by imputing it to a dummy message in the discourse and it looked fine there, not excluding that you still might be right or I am misunderstanding.

docs/src/models/biological.md
TorkelE: Ok, I think I got it to work:
$$
\frac{d[X]}{dt} = -2.0[X]\cdot [Y]\\
\frac{d[Y]}{dt} = -2.0[X]\cdot [Y]\\
\frac{d[XY]}{dt} = 2.0[X]\cdot [Y] - 1.0[XY]\\
\frac{d[Z1]}{dt}= 1.0[XY]\\
\frac{d[Z2]}{dt} = 1.0[XY]\\
$$
(very useful knowledge by the way :) )

Is there a way to update these pull requests without deleting and creating a new. Also, are your review automatically made as changes or do I have to make them happen to the PR?

ChrisRackauckas: it emails me of all changes so I can look at it. If you commit and push to the branch you opened the PR from it will update the PR.

TorkelE: Ok. I clicked on the "Edit on GitHub" link on the docs page (http://docs.juliadiffeq.org/latest/models/biological.html) and then the edit button and simply pasted the text in. I guess it is possible from there as well (although maybe I will try to do it the other way next time instead).

Either way, sending you the text.

korsbo: Your equation syntax is still off. Your example should be something like:
```math
\begin{align*}
\frac{\mathrm{d}\left[X\right]}{dt} =&  - 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] \\
\frac{\mathrm{d}\left[Y\right]}{dt} =&  - 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] \\
\frac{\mathrm{d}\left[XY\right]}{dt} =& 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] - \left[ XY \right] \\
\frac{\mathrm{d}\left[Z1\right]}{dt} =& XY \\
\frac{\mathrm{d}\left[Z2\right]}{dt} =& XY \\
\end{align*}
```
which is surrounded by ```math

docs/src/models/biological.md
```diff
@@ -1,166 +1,307 @@
...
+```
+Sometimes several molecules of the same reactant is involved in a reaction, the stoichiometry of a reactant in a reaction can be set using a number. Here two species of `X` forms the dimer `X2`:
+```julia
+rn = @reaction_network begin
+  1.0, 2X → X2
+end
+```
+this corresponds to the differential equation:
+d[X]/dt = -1.0\*[X\]^2/2!
+d[X2]/dt = 1.0\*[X\]^2/2!
```
korsbo:
```math
\begin{align*}
\frac{dX}{dt} =& -2 \cdot \frac{X^{2}}{2} \\
\frac{dX2}{dt} =& \frac{X^{2}}{2} \\
\end{align*}
```
surrounded by ```math

TorkelE: This PR is a bit outdated, but you seem to be right that some of these are missing.

How do you meant though, I try to check how
´´´
\begin{align*}
\frac{\mathrm{d}\left[X\right]}{dt} =&  - 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] \\
\frac{\mathrm{d}\left[Y\right]}{dt} =&  - 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] \\
\frac{\mathrm{d}\left[XY\right]}{dt} =& 2.0 \cdot \left[ X \right] \cdot \left[ Y \right] - \left[ XY \right] \\
\frac{\mathrm{d}\left[Z1\right]}{dt} =& XY \\
\frac{\mathrm{d}\left[Z2\right]}{dt} =& XY \\
\end{align*}
´´´
would look like (but with ´´´math before and simple three ´ after), but it does not look very LaTeXy?

(I try to preview it here and paste it into a discourse message and see how it looks, the only way I know how to actually look how .md formatting turns out like).

Looking at the big one it looks perfectly fine at
http://docs.juliadiffeq.org/latest/models/biological.html
(ignoring that the new lines does not work).
but on 
https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/master/docs/src/models/biological.md
it for some reason does not work at all. Checking on my discourse message it looks perfect. I have no idea why it works somewhere and somewhere not.

ChrisRackauckas:
>This PR is a bit outdated

? I think all that's needed to finish this is to tag the math sections appropriately? If so, I can just edit it and merge.

TorkelE:
I tried to edit this one, but was unable to (I think it was because I did not edit it on my computer, but went to http://docs.juliadiffeq.org/latest/models/biological.html, clicked "edit on github", clicked the edit file pen and then copied the new text in).

So instead I simply sent a new pull request with what I thought was correct math notation, which I think was pulled because http://docs.juliadiffeq.org/latest/models/biological.html uses the new docs.

However @korsbo is still right in that:
1) I forgot to use proper mathematical notation on 2 additional places.
2) The one currently used in http://docs.juliadiffeq.org/latest/models/biological.html does not have proper new lines.

However, this is the point where I got confused. Because I am a bit unsure how .md got rendered and wanted to check things I wrote everything in a discourse message and check the preview. When I checked the text which now can be found on http://docs.juliadiffeq.org/latest/models/biological.html on discourse it looked good. However, when I check it on:
1) http://docs.juliadiffeq.org/latest/models/biological.html there is no proper new lines.
2) https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/master/docs/src/models/biological.md it does not render at all.

If I understood @korsbo correctly I should write 
```math
\begin{align*}
\frac{dX}{dt} =& -2 \cdot \frac{X^{2}}{2} \\
\frac{dX2}{dt} =& \frac{X^{2}}{2} \\
\end{align*}
```
(but instead add math after the thrid `)
However, when I do that here it does not have any effect, neither does it have when I use my discourse "trick". 

Right now honestly have no idea what is happening, and why .md is rendered differently at different places. I hav tried to google how to do this, but found nothing really helpful.

korsbo:
Its all GitHubs fault. The GitHub webpages does not seem to support MathJax, so stuff does not get rendered. This is not an issue in the documentation webpage. The approach we have been advocating should work on the documentation webpage, even if it does not render in the GitHub preview.

TorkelE:
Cheers. I still only vaguely understand what is happening but the docs looks good at least.

ChrisRackauckas:
There's a few spots where the TeX seems to not have rendered properly, but for the most part it's good.

PR #144, opened 2018-08-09
Update ExpRK documentation

MSeeker1340:
Adds missing integrators and describes in detail the options for the ExpRK methods.

PR #150, opened 2018-08-21
Section: specifying jacobian prototype

MSeeker1340:
Completes http://docs.juliadiffeq.org/latest/features/performance_overloads.html#Specifying-Jacobian-Types-1.


MSeeker1340:
Also what about the updates here https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/144? Any further edits I need to make?

PR #164, opened 2018-10-18
Document tspan option for DynamicSS

tkf:
which is added in https://github.com/JuliaDiffEq/SteadyStateDiffEq.jl/pull/8


ChrisRackauckas:
Can the docstrings of algorithms be added like this to the docs? I don't want the gigantic box they put around them in most cases, but want to use docstrings.

tkf:
> gigantic box

You mean the gray border around the docstring like this?

image

I guess you can tweak the CSS to remove them. I've never tried though.

PR #165, opened 2018-10-19
Setup Documenter using Project.toml

tkf:
Implement:
https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974
https://discourse.julialang.org/t/psa-documenter-jl-breaking-changes-version-capping/16431

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.mem
 docs/build
 docs/site
+/Manifest.toml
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
looks like it still doesn't deploy since it needs debugging. I'll take a deeper look later, but might need a reminder.

tkf:
Did you already fix it? It seems the docs website is updated. [For example, it has `DiffEqBase.set_t!` docstring now.](http://docs.juliadiffeq.org/latest/basics/integrator.html#DiffEqBase.set_t!)

PR #167, opened 2018-10-19
Remove the box around docstrings

tkf:
It produces something like this:

image

See: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/371


tkf:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/371#issuecomment-431518210

PR #175, opened 2018-12-21
Add First difference and Two stage method examples

Vaibhavdixit02:


docs/src/analysis/parameter_estimation.md
```diff
@@ -545,7 +555,18 @@ Out[4]:2-element Array{Float64,1}:
...
         1.01721
 ```
-Here as our model had 2 parameters, we look at the last two indexes of `result` to get our parameter values and the rest of the values are the initial values of the shorter timespans as described in the reference section.
+Here as our model had 2 parameters, we look at the last two indexes of `result` to get our parameter values and 
+the rest of the values are the initial values of the shorter timespans as described in the reference section.
+
+The objective function for Two Stage method can be created as
+
+```julia
+two_stage_obj = two_stage_method(ms_prob,t,data)
```
ChrisRackauckas: optimize it?

PR #180, opened 2019-01-11
split local and global sensitivity and add information about AD

ChrisRackauckas:


docs/src/analysis/sensitivity.md
```diff
@@ -1,36 +1,158 @@
...
+1 in a unique index. For example, we can build our parameter vector like:
 
-## Local Sensitivity Analysis
+```julia
+using ForwardDiff: Dual
+p1dual = Dual{Float64}(1.5, (1.0, 0.0, 0.0, 0.0))
+p2dual = Dual{Float64}(1.0, (0.0, 1.0, 0.0, 0.0))
+p3dual = Dual{Float64}(3.0, (0.0, 0.0, 1.0, 0.0))
+p4dual = Dual{Float64}(3.0, (0.0, 0.0, 0.0, 0.0))
+pdual = [p1dual, p2dual, p3dual, p4dual]
```
ChrisRackauckas: @jrevels is there an easier way to seed Duals like this?

docs/src/analysis/sensitivity.md
```diff
@@ -274,7 +398,12 @@ to the internal ODE solver for solving the adjoint problem. Two special keyword
 arguments are `iabstol` and `ireltol` which are the tolerances for the internal
 quadrature via QuadGK for the resulting functional.
 
-### Example discrete adjoints on a cost function
+#### Options
+
+Options for handling the adjoint computation are set by passing a `SensitivityAlg`
```
ChrisRackauckas: @YingboMa can you fill this in?

ChrisRackauckas:
I'm purposely leaving out mentions of Flux and ReverseDiff here since they didn't benchmark well and this is already cluttered. We can add Zygote here when it's ready.

docs/src/analysis/sensitivity.md
jrevels: This is actually really clean from a pedagogical standpoint. There is also [this constructor](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/dual.jl#L67), which allows you to write, for example:

```julia
using ForwardDiff: Dual, Chunk
chunk = Chunk{4}()
pdual = [Dual(1.5, chunk, Val(1)),
         Dual(1.0, chunk, Val(2)),
         Dual(3.0, chunk, Val(3)),
         Dual(3.0, chunk, Val(4))]
```

If you want to do it programmatically like ForwardDiff does it, see, for example, https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/apiutils.jl#L17 and https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/config.jl#L117.

BTW, the constructor is `Dual{T}(...)` where `T` is the tag type, not the element type (I'm assuming you don't actually want to use a `Float64` tag).

ChrisRackauckas: I thought about writing out the programmatic way, but IMO that should be upstreamed to a convenience function in ForwardDiff. PR: https://github.com/JuliaDiff/ForwardDiff.jl/pull/379

PR #181, opened 2019-01-14
Add options of `SensitivityAlg`

YingboMa:


docs/src/analysis/sensitivity.md
```diff
@@ -408,7 +408,21 @@ quadrature via QuadGK for the resulting functional.
 #### Options
 
 Options for handling the adjoint computation are set by passing a `SensitivityAlg`
-type.
+type, e.g. `SensitivityAlg(backsolve=true)`.
```
ChrisRackauckas: We should show a real call here

PR #186, opened 2019-01-22
Fix typos

jonas-schulze:
to be consistent with [the ODE example tutorial](https://docs.juliadiffeq.org/latest/tutorials/ode_example.html#Choosing-a-Solver-Algorithm-1).


jonas-schulze:
Why is there a separate repository for documenting `DifferentialEquations.jl`? As that one has no `docs/` folder, maybe one could think of integrating this one into the other.

ChrisRackauckas:
>Why is there a separate repository for documenting DifferentialEquations.jl? 

It used to be there, but it blew up the size of the repository so it had to be moved.

ChrisRackauckas:
Thanks!

Issue #184, opened 2019-01-21
Wrong bulleting in homepage


kanav99:
In the homepage, in the index of Solver Algorithms, we have something like this:

* BVP Solvers
    * Recommended Methods
* Full List of Methods
    * BoundaryValueDiffEq.jl

Though I think it should be:

* BVP Solvers
    * Recommended Methods
    * Full List of Methods
        * BoundaryValueDiffEq.jl

ChrisRackauckas:
Do we have a third nest in the docs? Anyways, changing this is just changing the depth of the header.

PR #188, opened 2019-01-23
Fixes Indenting

kanav99:
Addressed in #184 , this fixes the issue with depth of 2. We can either have a depth of 2:
d2

or we can have depth of 3:
d3

Currently the depth is set to 2. We can change it accordingly.


ChrisRackauckas:
Looks nice, thanks!

PR #195, opened 2019-02-08
Added new methods

deeepeshthakur:
@ChrisRackauckas Docs for new DGLDDRK methods.
`DGLDDRK73_C`
`DGLDDRK84_C`
`DGLDDRK84_F`

### DIFF
--- a/docs/src/solvers/ode_solve.md
+++ b/docs/src/solvers/ode_solve.md
@@ -275,6 +275,8 @@ Royal Society, 2011.).
   low-dispersion scheme for discontinuous Galerkin space discretizations
   applied to wave propagation problems, optimized for PDE discretizations
   when the maximum spatial step size is not constrained. Fixed timestep only.
+- `HSLDDRK64` - 6-stage, fourth order low-stage, low-dissipation, low-dispersion
+  scheme. Fixed timestep only.
 - `RK46NL` - 6-stage, fourth order low-stage, low-dissipation, low-dispersion
   scheme. Fixed timestep only.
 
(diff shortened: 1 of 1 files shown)


deeepeshthakur:
@ChrisRackauckas Should I add the difference between _C and _F methods?
`_C` is optimized for scenarios where maximum step size is constrained by geometrical features.
`_F` is optimized for scenarios where maximum step size is free.

ChrisRackauckas:
>`_C` is optimized for scenarios where maximum step size is constrained by geometrical features.
>`_F` is optimized for scenarios where maximum step size is free.

That needs a bit more explanation. How do they explain it in the papers?

deeepeshthakur:
They described it like this:
```
From a user point of view, the problem of the computational efficiency involves 
the choice of the best combination of mesh and numerical method; two 
scenarios are defined. In the first one, the element size is totally free, 
and a 8-stage, fourth-order Runge–Kutta scheme is found to minimize a 
cost measure depending on both accuracy and stability. In the second one, 
the elements are assumed to be constrained to such a small size by 
geometrical features of the computational domain, that accuracy is disregarded. 
We then derive one 7-stage, third-order scheme and one 8-stage, fourth-order 
scheme that maximize the stability limit.
```

ChrisRackauckas:
Your description makes no reference to the fact that what is constrained is not the time step but the spatial step. `_F` is optimized for PDE discretizations with free element sizes, and `_C` is for when the element sizes can get small due to geometric features of the domain.

deeepeshthakur:
@ChrisRackauckas  I'll change it.

deeepeshthakur:
And I think we missed the LDDRK64 methods (now HSLDDRK64), so I've added it.

PR #202, opened 2019-02-26
document assumed reaction rate scaling used in DiffEqBiological

isaacsas:


ChrisRackauckas:
hmm I thought we had this in there, but I guess not!

isaacsas:
Yeah, that was definitely an oversight... (Unless I missed where we had it.)

It bit me when testing the BioNetGen parser, as they already rescale the rates by those factors. I eventually realized I had to correct for that when I was comparing their ODE solutions to what I was getting in Julia with the generated reaction network...

ChrisRackauckas:
Wait, don't we auto-rescale? So shouldn't BioNetGen be outputting the same rates as us?

isaacsas:
We auto-rescale, but the .net format I'm parsing from BioNetGen is a greatly simplified format that is auto-generated from their BioNetGen Language (bngl). bngl is much more compact, human writable format, but much too complicated for me to tackle parsing. In the bngl file the rates follow the same convention we use, but the auto-generated .net file seems to already include the rescalings, hence I need to correct for them.

isaacsas:
Actually I misremembered -- in the bngl files they also state that one should explicitly input a scaled rate when inputting a single reaction, i.e. for A+A->B the rate should be input as .5*k, where k is the rate constant we would input.

PR #204, opened 2019-03-05
Fix some Tex

Vaibhavdixit02:


ChrisRackauckas:
looks like only one of the tex issues?

PR #206, opened 2019-03-07
Add KuttaPRK2p5

saurabhkgp21:


docs/src/solvers/ode_solve.md
```diff
@@ -159,6 +159,7 @@ problems.
 - `Feagin10` - Feagin's 10th-order Runge-Kutta method.
 - `Feagin12` - Feagin's 12th-order Runge-Kutta method.
 - `Feagin14` - Feagin's 14th-order Runge-Kutta method.
+- `KuttaPRK2p5` - 5th order 2 processor parallel across the method Runge Kutta method.
```
ChrisRackauckas: Make them a new section and describe their use case in more detail

PR #208, opened 2019-03-13
DiffEqBiological API

isaacsas:
This adds a new sidebar category for APIs, with a page containing a generated API for `DiffEqBiological` based on the comments I've added there. On my computer the webpages generate fine and look good.

Note, this make DiffEqDocs depend on DiffEqBiological. As the later needs Julia 1.1 I bumped REQUIRE and travis to 1.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: julia
 os:
   - linux
 julia:
-  - 1.0
+  - 1.1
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Yeah, thanks for starting us off here. We need to get this going for all packages.

PR #218, opened 2019-03-29
Update sde_example.md

GantZA:
Fixed typo, I think, t₀ to u₀


ChrisRackauckas:
And I updated the analytical solution syntax as well.

ChrisRackauckas:
Thanks!

PR #219, opened 2019-04-03
changed sparse(A) by A=sparse(A)

WouterJRV:
sparse!(A) does not seem to exist...

### DIFF
--- a/docs/src/tutorials/sde_example.md
+++ b/docs/src/tutorials/sde_example.md
@@ -291,7 +291,7 @@ A = zeros(2,4)
 A[1,1] = 1
 A[1,4] = 1
 A[2,4] = 1
-sparse(A)
+A=sparse(A)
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Thanks!

WouterJRV:
It's not completely the end of the story it seems, sparse to my particular problem still gives the error though I haven't succeeded so far in making the test problem fail as well.

ChrisRackauckas:
Has this been tracked down?

WouterJRV:
> Has this been tracked down?

It seems something subtle, I've been trying to give the test problem specified noise, complex contributions, zeros in A, time-adaptive methods and a number of different options without problems

For the problems with which there is this error, I see that the error "no method matching zeros(Type::any)" vanishes either if A is not sparse, or if I put reltol=Inf.

PR #221, opened 2019-04-28
Documentation for explicit extrapolation methods

AlthausKonstantin:
Added documentation for ExtrapolationMidpointDeuflhard and ExtrapolationMidpointHairerWanner (cf. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/714)

### DIFF
--- a/docs/src/solvers/ode_solve.md
+++ b/docs/src/solvers/ode_solve.md
@@ -325,94 +325,37 @@ So the above implementation of `f` becomes valid.
 
 #### Extrapolation Methods
 
-- `AitkenNevillie` - An adaptive order, adaptive step size extrapolation method using Aitken-Neville 
-  algorithm. The step-number sequence is the Romberg Sequence.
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This is a bit wordy and convoluted. Most of these details aren't needed for the docs and should be in the PR instead.

ChrisRackauckas:
I simplified it a bit and put a commit to OrdinaryDiffEq to unify the arg naming.

ChrisRackauckas:
The non-stiff ODE benchmarks from https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/pull/32 are updated. You might want to make a PR to that branch and see how the extrapolation methods fare on some of those benchmarks.

AlthausKonstantin:
All right, didn't know that I would have gotten away with such brevity. Thanks for going over it!
I will have a look at that PR.

PR #226, opened 2019-06-05
Adds VectorContinuousCallback documentation

kanav99:
I have added documentation for now, will add an example of bouncing ball with two walls as an example.


ChrisRackauckas:
Examples can come next.

PR #227, opened 2019-06-06
Adds VectorContinuousCallback example

kanav99:
Graph looks like this: 
ball2


ChrisRackauckas:
Can you add this example to the release blog post?

kanav99:
Yeah for sure! Should I explain everything there too or would the code and graph suffice?

ChrisRackauckas:
Code, graph, and link to docs. Basically just say, look at a ball bounce off of two walls.

PR #232, opened 2019-06-24
mentioned QuDiffEq.jl

dgan181:


ChrisRackauckas:
It should go in the ODE solver page instead.

dgan181:
Yeah. Sorry. I didn't look at the message carefully. So is this fine or should I add an example?

ChrisRackauckas:
Take a look at this page. Make a section after this one http://docs.juliadiffeq.org/latest/solvers/ode_solve.html#TaylorIntegration.jl-1 that introduces the new algorithm choices and its options.

dgan181:
Ah. So I just include the last method (the one that uses `::ODEProblem`) then? 
The linearised form for HHL can be developed just as simply as for the other one. I was planning to add that right after fixing some implementation issues in HHL. I was having trouble with the getting the correct order of precision in higher order methods. I still need to figure that out.
FYI,  HHL isn't a great way to go about solving differential equations because circuit becomes increasingly bulky with more dimensions (variables and time steps).

ChrisRackauckas:
>Ah. So I just include the last method (the one that uses ::ODEProblem) then?

Yup, or add the HHL ones if they are going to be done soon.

docs/src/solvers/ode_solve.md
```diff
@@ -1047,6 +1047,22 @@ using TaylorIntegration
 
 Note: this method is much faster if you put `@taylorize` on your derivative function!
 
+  ### QuDiffEq.jl
```
ChrisRackauckas:
```suggestion
### QuDiffEq.jl
```

PR #233, opened 2019-06-24
Docs for ODE and SDE solvers

deeepeshthakur:
@ChrisRackauckas

### DIFF
--- a/docs/src/solvers/sde_solve.md
+++ b/docs/src/solvers/sde_solve.md
@@ -169,7 +169,7 @@ For `SRA` and `SRI`, the following option is allowed:
 - `SKSROCK` - is fixed step stabilized explicit method for stiff Ito problems. Strong order 0.5
   and weak order 1. This method has a better stability domain then `SROCK1`. Also it allows
   special post-processing techniques in case of ergodic dynamical systems, in the context of
-  ergodic Brownian dynamics, to achieve order 2 accuracy. `SKSROCKSKSROCK(;post_processing=true)`
+  ergodic Brownian dynamics, to achieve order 2 accuracy. `SKSROCK(;post_processing=true)`
   will make use of post processing. By default it doesn't use post processing. Post processing is
   optional and under development. The rest of the method is completely functional and can handle
   1-dimensional, diagonal and multi-dimensional noise.  
--- a/docs/src/solvers/sde_solve.md
+++ b/docs/src/solvers/sde_solve.md
@@ -162,19 +162,20 @@ For `SRA` and `SRI`, the following option is allowed:
   optimised to get order 1 in case os scalar/diagonal noise.
 - `SROCKEM` - is fixed step Euler-Mayurama with first order ROCK stabilization thus can
   handle stiff problems. Only for Ito problems. Defaults to strong and weak order 1.0,
-  but can solve with weak order 0.5 as `SROCKEM(strong_order_1=false)`.
+  but can solve with weak order 0.5 as `SROCKEM(strong_order_1=false)`. This method can handle
+  1-dimensional, diagonal and multi-dimensional noise.
 - `SROCK2` - is a weak second order and strong first order fixed step stabilized method for
-  stiff Ito problems.
+  stiff Ito problems.This method can handle 1-dimensional, diagonal and multi-dimensional noise.
 - `SKSROCK` - is fixed step stabilized explicit method for stiff Ito problems. Strong order 0.5
   and weak order 1. This method has a better stability domain then `SROCK1`. Also it allows
   special post-processing techniques in case of ergodic dynamical systems, in the context of
-  ergodic Brownian dynamics, to achieve order 2 accuracy.
+  ergodic Brownian dynamics, to achieve order 2 accuracy. `SKSROCKSKSROCK(;post_processing=true)`
+  will make use of post processing. By default it doesn't use post processing. Post processing is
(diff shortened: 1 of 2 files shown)


docs/src/solvers/sde_solve.md
```diff
@@ -111,6 +111,12 @@ Orders are given in terms of strong order.
   Milstein method for commutative noise problems. Defaults to solving the Ito
   problem, but `RKMilCommute(interpretation=:Stratonovich)` makes it solve the
   Stratonovich problem. Uses a 1.5/2.0 error estimate for adaptive time stepping.†
+- `WangLi3SMil_A` - fixed step-size explicit 3-stage Milstein methods for Ito problem with strong and weak order 1.0
```
ChrisRackauckas: Mention the noise that's allowed.

docs/src/solvers/sde_solve.md
```diff
@@ -147,6 +153,18 @@ For `SRA` and `SRI`, the following option is allowed:
...
+#### S-ROCK Methods
+
+- `SROCK1` - is a fixed step size stabilized explicit method for stiff problems. Defaults to
+  solving th Ito problem but `SROCK1(interpretation=:Stratonovich)` can make it solve
+  the Stratonovich problem. Strong order of convergence is 0.5 and weak order 1, but is
+  optimised to get order 1 in case os scalar/diagonal noise.
+- `SROCKEM` - is fixed step Euler-Mayurama with first order ROCK stabilization thus can
+  handle stiff problems. Only for Ito problems. Defaults to strong and weak order 1.0,
+  but can solve with weak order 0.5 as `SROCKEM(strong_order_1=false)`.
+- `SROCK2` - is a weak second order and strong first order fixed step stabilized method for
```
ChrisRackauckas: isn't there more?

deeepeshthakur: But they are not fully functional.

ChrisRackauckas: Add a note about that in the description. Then we can delete it when they are done. It's better to just document all we can while we're thinking about it.

deeepeshthakur: Added!

deeepeshthakur:
outofplace vesrion will give error in matrix multiplication for diagonal noise

ChrisRackauckas:
there is no matmul for diagonal noise? Just a broadcast?

deeepeshthakur:
Yes but when I was writing those. I only had in mind 1D noise. So I didn’t use broadcasting in outofplace version, So if we have something like a 4x3 dW it would be a problem.

deeepeshthakur:
I’ll fix it today and then we can go forward with this PR

ChrisRackauckas:
The diagonal case is just W is a vector

deeepeshthakur:
Yes. But what I'm saying that we have these type of equations so they won't work.
```
k  = integrator.g(u,p,t)
u  = u + k*W.dW + (W.dW^2/2)*k₁
```

ChrisRackauckas:
Oh okay, it would if you `@..`

deeepeshthakur:
Yeah I'll fix it today.

docs/src/solvers/sde_solve.md
```diff
@@ -140,13 +146,36 @@ Example usage:
...
+  stiff Ito problems.
+- `SKSROCK` - is fixed step stabilized explicit method for stiff Ito problems. Strong order 0.5
+  and weak order 1. This method has a better stability domain then `SROCK1`. Also it allows
+  special post-processing techniques in case of ergodic dynamical systems, in the context of
+  ergodic Brownian dynamics, to achieve order 2 accuracy.
+- `TangXiaoSROCK2` - is a fixed step size stabilized expicit method for stiff problems. Only for
+  Ito problems. Weak order of 2 and strog order of 1. Has 5 versions with different stability
+  domains which can be used as `TangXiaoSROCK2(version_num=i)` where `i` is 1-5.
+
+Note: Post processing in `SKSROCK` is still under development. Also `TangXiaoSROCK2` is also not fully
```
ChrisRackauckas: but post processing is an optional argument which is off by default IIRC? You need to document the argument, and mention that it's just fine without the post processing.

docs/src/solvers/sde_solve.md
```diff
@@ -140,13 +146,36 @@ Example usage:
...
+  but can solve with weak order 0.5 as `SROCKEM(strong_order_1=false)`.
+- `SROCK2` - is a weak second order and strong first order fixed step stabilized method for
+  stiff Ito problems.
+- `SKSROCK` - is fixed step stabilized explicit method for stiff Ito problems. Strong order 0.5
+  and weak order 1. This method has a better stability domain then `SROCK1`. Also it allows
+  special post-processing techniques in case of ergodic dynamical systems, in the context of
+  ergodic Brownian dynamics, to achieve order 2 accuracy.
+- `TangXiaoSROCK2` - is a fixed step size stabilized expicit method for stiff problems. Only for
+  Ito problems. Weak order of 2 and strog order of 1. Has 5 versions with different stability
+  domains which can be used as `TangXiaoSROCK2(version_num=i)` where `i` is 1-5.
```
ChrisRackauckas:
```suggestion
  domains which can be used as `TangXiaoSROCK2(version_num=i)` where `i` is 1-5. Under Development.
```

deeepeshthakur:
Fixed. And also I've added diagonal noise in docs

PR #236, opened 2019-07-09
Add kwargs to linsolve documentation

Tyde:
See also here: https://stackoverflow.com/questions/54545738/solve-the-heat-equation-with-non-zero-dirichlet-bcs-with-implicit-euler-and-conj/54545739#comment100407633_54545739


ChrisRackauckas:
Thanks!

PR #262, opened 2019-08-04
Add Rosenbrock-W method

huanglangwen:


docs/src/solvers/ode_solve.md
```diff
@@ -493,6 +493,11 @@ These methods also have option `nlsolve` same as SDIRK methods. These methods al
 - `GRK4A` - An A-stable 4th order Rosenbrock method. Essentially "anti-L-stable"
   but efficient.
 - `Ros4LStab` - A 4th order L-stable Rosenbrock method.
+- `RosenbrockW6S4OS` - A 4th order L-stable Rosenbrock-W method (fixed step only).
+- `ROS34PW1a` - A 4th order L-stable Rosenbrock-W method.
```
ChrisRackauckas: These support mass matrices too, right? That should get mentioned in the DAE page (and a test)

huanglangwen: Yes, but out-of-place methods don't support it, just like others.

ChrisRackauckas: Add it to the DAE page, and we'll fix up the out of place ones. I am hoping this gets done this week BTW, either me, you, @YingboMa , or @ranjanan should probably take a look at it.

huanglangwen: I can fix the generic one.

ChrisRackauckas: IIRC what needs to be fixed is the oop W generation and that's it? So if you fix one you fix all.

huanglangwen: Not really, the oop stepping code ignored mul of mass_matrix on du.
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/abaf387213e1b1b020b795242cf4bd5ca2a950d8/src/perform_step/rosenbrock_perform_step.jl#L129-L135
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/abaf387213e1b1b020b795242cf4bd5ca2a950d8/src/perform_step/rosenbrock_perform_step.jl#L217

ChrisRackauckas: that doesn't need to be modified since it's in the W matrix.  W = M - gamma*J instead of W = I - gamma*J to generalize it, and it uses the same linear solve.

huanglangwen: But why inplace code have that?

ChrisRackauckas: Oh right, I forgot that part lol

docs/src/solvers/dae_solve.md
```diff
@@ -26,6 +26,18 @@ extra options for the solvers, see the ODE solver page.
 - `ROS3P` - 3rd order A-stable and stiffly stable Rosenbrock method. Keeps high
   accuracy on discretizations of nonlinear parabolic PDEs.
 - `Rodas3` - 3rd order A-stable and stiffly stable Rosenbrock method.
+- `RosShamp4`- An A-stable 4th order Rosenbrock method.
+- `Veldd4` - A 4th order D-stable Rosenbrock method.
+- `Velds4` - A 4th order A-stable Rosenbrock method.
```
ChrisRackauckas: Let's make sure these others added to the mass matrix tests @YingboMa

huanglangwen: Luckily, ROS4 and ROSW are included in the macros. We only need to check `Rosenbrock23/32` class and `Rodas` class.

ChrisRackauckas: Those already have tests. The other ones, like RosShamp4, I don't think are in the mass matrix tests

huanglangwen: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/862

ChrisRackauckas: Awesome. It looks like Rosenbrock23 should be added to this list as well

ChrisRackauckas:
The Rosenbrock sections are getting long. should we separate the W methods out to a separate section?

huanglangwen:
> 
> 
> All alg section headers should have 4

But they are under "Methods for Stiff Problems" section:

```markdown
#### Methods for Stiff Equations
##### SDIRK Methods
##### Parallel Diagonally Implicit Runge-Kutta Methods
##### Fully-Implicit Runge-Kutta Methods (FIRK)
##### Rosenbrock Methods
##### Rosenbrock-W Methods
##### Stabilized Explicit Methods
##### Exponential Methods for Linear and Affine Problems
```

ChrisRackauckas:
Your branch is just old: https://raw.githubusercontent.com/JuliaDiffEq/DiffEqDocs.jl/master/docs/src/solvers/ode_solve.md ?

Issue #272, opened 2019-09-03
Solving an ODE With Different Initial Conditions example does not work


logankilpatrick:
I have the following code per the example [here](http://docs.juliadiffeq.org/latest/features/ensemble.html#Example-1:-Solving-an-ODE-With-Different-Initial-Conditions-1):

```julia 
using Distributed
using DifferentialEquations
using Plots

addprocs()
@everywhere using DifferentialEquations

# Linear ODE which starts at 0.5 and solves from t=0.0 to t=1.0
prob = ODEProblem((u,p,t)->1.01u,0.5,(0.0,1.0))

@everywhere function prob_func(prob,i,repeat)
  ODEProblem(prob.f,rand()*prob.u0,prob.tspan)
end

ensemble_prob = EnsembleProblem(prob,prob_func=prob_func)
sim = solve(ensemble_prob,Tsit5(),trajectories=100)

plotly()
plot(sim,linealpha=0.4)
```

__I get the following error:__
```julia 
ERROR: LoadError: UndefKeywordError: keyword argument num_monte not assigned
Stacktrace:
 [1] (::getfield(DiffEqBase, Symbol("#kw##__solve")))(::NamedTuple{(:trajectories,),Tuple{Int64}}, ::typeof(DiffEqBase.__solve), ::EnsembleProblem{ODEProblem{Float64,Tuple{Float64,Float64},false,Nothing,ODEFunction{false,getfield(Main, Symbol("##9#10")),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},Nothing,DiffEqBase.StandardODEProblem},typeof(prob_func),getfield(DiffEqBase, Symbol("##293#299")),getfield(DiffEqBase, Symbol("##295#301")),Array{Any,1}}, ::Tsit5, ::MonteSerial) at ./none:0
[...]

PR #274, opened 2019-09-03
Update ensemble.md

logankilpatrick:
Added note about having the correct Julia version.

### DIFF
--- a/docs/src/features/ensemble.md
+++ b/docs/src/features/ensemble.md
@@ -245,6 +245,8 @@ around the mean.
 
 ## Example 1: Solving an ODE With Different Initial Conditions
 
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Will close #272

ChrisRackauckas:
I think it's assumed no one is on 1.0 anymore. There is a lot more that requires that

logankilpatrick:
Guess I'm in the minority running 1.0.3/1.0.4 :) 

On the contrary, it's the "Long Term Supported Release"

ChrisRackauckas:
I think we'll just punt on this and hope that the long term stable release is soon something that is like 1.4, otherwise we'd need to write this on half of our docs 🤷‍♂ .

PR #275, opened 2019-09-07
Fix broken link to gsu

jarlebring:
rclewley's page on gsu has moved. I think this is an equivalent page.


ChrisRackauckas:
Thanks!

PR #277, opened 2019-09-28
Add favicon for docs

Ayushk4:
http://docs.juliadiffeq.org/ did not have a favicon - 

now

This favicon has been taken from https://github.com/JuliaDiffEq/juliadiffeq.github.io/blob/master/favicon.png

After building the current patch locally -

new


ChrisRackauckas:
Looks like it failed to build? Maybe it's the version of Documenter?

Ayushk4:
I am looking into it.

Ayushk4:
> Maybe it's the version of Documenter?
Yes. This repo uses 19.x vs the one locally used of 23.x. I have changed syntax back to 19.x for the time being.

Let me know if this is okay or should I test it against v23.x and make changes accordingly

ChrisRackauckas:
This is fine. Updating the docs would be a good idea sometime, but one step at a time 👍

PR #280, opened 2019-10-17
Proposing an example of an nonhomogeneous ODE for the tutorial

hurak:
I propose inserting an example of a **nonhomogeneous ODE** into the tutorial. Questions about how to build an ODEProblem for such equations (also referred to as ODEs with **nonzero right-hand sides**) appear every now and then (I guess they mostly come from engineers who are viewing such equations as models of systems with **inputs**). The example illustrates how the framework of parameterized functions can be used for accomplishing this goal.

The example has even a simple physical interpretation: a slender rod pendulum exposed to an external torque.


docs/src/tutorials/ode_example.md
```diff
@@ -391,7 +391,48 @@ can also be used for sensitivity analysis, parameter estimation routines,
...
+
+prob = ODEProblem(pendulum!,u₀,tspan,M)
+sol = solve(prob)
+
+plot(sol,linewidth=2,xaxis="t",label=["θ [rad]" "ω [rad/s]"],layout=(2,1))
+```
+
+![Pendulum response](../assets/pendulum_response.png)
+
+In particular, note how the external torque `M` is introduced as a *time-varying* parameter in the `pendulum!` function. Indeed, unlike the (vector of) state variables `u`, it must have the dependence on time explicitly expressed.
```
ChrisRackauckas: we might want to describe this instead as a general principle that the parameters can be any type. Here, we specify `M` as time-varying by representing it by a function.

ChrisRackauckas:
I think this is a great idea!

docs/src/tutorials/ode_example.md
hurak: Perhaps the paragraph should be modified into something like this:

Note how the external **time-varying** torque `M` is introduced as a **parameter** in the `pendulum!` function. Indeed, as a general principle the parameters can be any type; here we specify `M` as time-varying by representing it by a function, which is expressed by appending the dependence on time `(t)` to the name of the parameter.  

Note also that, in contrast with the time-varying parameter, the (vector of) state variables `u`, which is generally also time-varying, is always used without the explicit dependence on time `(t)`.

ChrisRackauckas: yeah, that sounds good!

ChrisRackauckas:
Thanks!

PR #284, opened 2019-10-25
API documentation skeleton for DiffEqBase

jlumpe:
Created a skeleton for the `DiffEqBase` API documentation. I haven't actually written much content, but there are docstrings added for nearly all types and most functions categorized into different pages and subheadings. I'm hoping having this framework in place will make it easy for people who know the codebase better than I to start filling in the rest.

I created a PR to `DiffEqBase` that added docstring stubs for many types and functions that were lacking them, so until that's accepted you'll get a lot of "no docs found for ..." warnings.


jlumpe:
Also, seems like #285 will have to be merged in order to install the most recent version of `DiffEqBase` (or anything past 3.0.0).

PR #292, opened 2019-11-05
time no longer needs to be dual type

ArnoStrouwen:
Both ForwardDiff.jl and Calculus.jl give me the same answer.
But this answer is slightly different from the result previously in the docs, which was also the same for ForwardDiff.jl and Calculus.jl.
Was this example always wrong?


ChrisRackauckas:
I may have generated it at a different tolerance.

PR #297, opened 2019-11-17
Explain sensitivity options in more detail

ChrisRackauckas:


ChrisRackauckas:
Going to merge because it's at least better, though opinions from @YingboMa and @jlumpe could be helpful in further improving this section of the documentation.

PR #299, opened 2019-11-20
Add redirect for old links

ChrisRackauckas:
@mortenpi like this?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -122,7 +122,7 @@ cd(joinpath(@__DIR__, "build")) do
             isfile(redirect) && (@warn "$redirect exists, skip"; continue)
             open(redirect, "w") do io
                 write(io, """
-                <meta http-equiv="refresh" content="0; url=$(basename(m[1]))/index.html"/>
+                <meta http-equiv="refresh" content="0; url=$(basename(m[1]))/"/>
                 """)
             end
         end
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -113,3 +113,22 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological],
 deploydocs(
    repo = "github.com/JuliaDiffEq/DiffEqDocs.jl.git"
 )
+
+#Redirect old links
+cd("build/") do
+    for (root, dirs, files) in walkdir(".")
+        for file in files
+            path = relpath(joinpath(root, file), ".")
+            m = match(r"(.+)/index\.html$", path)
+            m === nothing && continue
+            redirect = "$(m[1]).html"
+            @info "Adding redirect for $(m[1]) from $(redirect)"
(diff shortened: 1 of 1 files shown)


docs/make.jl
```diff
@@ -113,3 +113,22 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological],
 deploydocs(
    repo = "github.com/JuliaDiffEq/DiffEqDocs.jl.git"
 )
+
+#Redirect old links
+cd("build/") do
```
fredrikekre: Since you are not running in the docs directory:
```suggestion
cd(@__DIR__) do
```

docs/make.jl
```diff
@@ -113,3 +113,22 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological],
 deploydocs(
    repo = "github.com/JuliaDiffEq/DiffEqDocs.jl.git"
 )
+
+#Redirect old links
```
fredrikekre: You need to do this between `makedocs` and `deploydocs`.

ChrisRackauckas:
@fredrikekre all good now?

docs/make.jl
```diff
@@ -110,6 +110,25 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological],
          ]
          ])
 
+#Redirect old links
+cd(@__DIR__) do
```
mortenpi: The `.` in `walkdir` should refer to `build/`:

```suggestion
cd(joinpath(@__DIR__, "build")) do
```

Otherwise it might do stuff in `src/` etc. too.

docs/make.jl
```diff
@@ -110,6 +110,25 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological],
...
+        for file in files
+            path = relpath(joinpath(root, file), ".")
+            m = match(r"(.+)/index\.html$", path)
+            m === nothing && continue
+            redirect = "$(m[1]).html"
+            @info "Adding redirect for $(m[1]) from $(redirect)"
+            isfile(redirect) && (@warn "$redirect exists, skip"; continue)
+            open(redirect, "w") do io
+                write(io, """
+                <meta http-equiv="refresh" content="0; url=$(basename(m[1]))/index.html"/>
```
fredrikekre:
```suggestion
                <meta http-equiv="refresh" content="0; url=$(basename(m[1]))/"/>
```

mortenpi: Fair point!

ChrisRackauckas:
All good now?

mortenpi:
LGTM, but only one way to find out for sure :laughing:

ChrisRackauckas:
yup, living on the wild side. glhf.

PR #306, opened 2019-11-24
The Wiener process is not white noise

eviatarbach:
I believe there is a mistake here, the Wiener process is not white noise, but rather Brownian (or red) noise. The increments are white Gaussian noise, but the process is not. See [the Wikipedia article](https://en.wikipedia.org/wiki/Wiener_process).


ChrisRackauckas:
Indeed

Issue #15, opened 2017-03-18
Add param_values to ParameterizedFunctions docs


ChrisRackauckas:
ParameterizedFunctions are getting a more in-depth interface, and it needs to start getting documented

PR #307, opened 2019-11-28
start structural identification documentation

ChrisRackauckas:
Most of this is complete, some of it is aspirational and brings up interface questions.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -86,6 +86,7 @@ makedocs(modules=[DiffEqBase,DiffEqProblemLibrary,DiffEqBiological,ODEProblemLib
          "Analysis Tools" => Any[
            "analysis/parameterized_functions.md",
            "analysis/parameter_estimation.md",
+           "analysis/structural_estimation.md",
            "analysis/bifurcation.md",
            "analysis/sensitivity.md",
            "analysis/global_sensitivity.md",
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@AlCap23 let's talk through a few things here:

- Default dt of eDMD vs ExactDMD: why 1.0 vs 0.0?
- `update!(dmd, x; Δt = 0.0, threshold = 1e-3)` should be possible. Is there a reason to ever require the user to split the data in the input functions?
- Latexify outputs from SInDy, i.e. https://github.com/JuliaDiffEq/DataDrivenDiffEq.jl/issues/17 . I think this would just be super cool.

AlCap23:
> Default dt of eDMD vs ExactDMD: why 1.0 vs 0.0?

As mentioned in [#15](https://github.com/JuliaDiffEq/DataDrivenDiffEq.jl/issues/15) there was no specific reason for the dt . 

> update!(dmd, x; Δt = 0.0, threshold = 1e-3) should be possible. Is there a reason to ever require the user to split the data in the input functions?

Sounds good. At the time that was not a priority. 

>Latexify outputs from SInDy, i.e. JuliaDiffEq/DataDrivenDiffEq.jl#17 . I think this would just be super cool.

Started working on that.

I will convert the example code with weave to jmd files and maybe use some nice examples out of papers.

PR #309, opened 2019-12-06
WIP: Documentation Overhaul

kanav99:

PR #310, opened 2019-12-09
Julia Programming Documentation Updates

wilsonhj:
- Changed all absolute links in /docs/src to relative links (@ref)
- Changed all relative file paths to relative links (@ref)
-DID NOT change any relative paths that point to a .png, i.e 
_Better Automatic Solution_
because I am not yet sure whether it is necessary to include the file extension .png, i.e
 _Better Automatic Solution_

Hope this helps!

### DIFF
--- a/docs/src/analysis/bifurcation.md
+++ b/docs/src/analysis/bifurcation.md
@@ -7,7 +7,7 @@ Bifurcation analysis on DifferentialEquations.jl types can be performed by:
 - PyDSTool.jl
 - DiffEqBiological.jl
 
-If your system is a chemical reaction system, see the documentation at 
-[DiffEqBiological.jl](https://github.com/JuliaDiffEq/DiffEqBiological.jl#making-bifurcation-diagram)
+If your system is a chemical reaction system, see the documentation at
+[DiffEqBiological.jl](@ref)
 for quickly generating bifurcation plots. Bifurcations.jl can directly generate a `BifurcationProblem`
 from an `ODEProblem`. PyDSTool.jl is no longer recommended.
--- a/docs/src/types/rode_types.md
+++ b/docs/src/types/rode_types.md
@@ -30,7 +30,7 @@ if you set a `callback` in the problem, then that `callback` will be added in
 every solve call.
 
 For specifying Jacobians and mass matrices, see the
-[DiffEqFunctions](http://docs.juliadiffeq.org/dev/features/performance_overloads)
+[DiffEqFunctions](@ref)
 page.
 
 ### Fields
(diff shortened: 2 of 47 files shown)


docs/src/analysis/bifurcation.md
```diff
@@ -7,7 +7,7 @@ Bifurcation analysis on DifferentialEquations.jl types can be performed by:
 - PyDSTool.jl
 - DiffEqBiological.jl
 
-If your system is a chemical reaction system, see the documentation at 
-[DiffEqBiological.jl](https://github.com/JuliaDiffEq/DiffEqBiological.jl#making-bifurcation-diagram)
+If your system is a chemical reaction system, see the documentation at
+[DiffEqBiological.jl](@ref)
```
ChrisRackauckas: this is not correct: `@ref` doesn't work for external links.

docs/src/types/rode_types.md
```diff
@@ -41,7 +41,7 @@ page.
 * `p`: The optional parameters for the problem. Defaults to `NullParameters`.
 * `noise`: The noise process applied to the noise upon generation. Defaults to
   Gaussian white noise. For information on defining different noise processes,
-  see [the noise process documentation page](../../../features/noise_process)
+  see [the noise process documentation page](@ref noise_process)
```
ChrisRackauckas: This is only correct if a label is put on the page, or it matches the title exactly.

ChrisRackauckas:
The vast majority of these changes are incorrect. The two repeated issues are found in the comments above. I would recommend trying to build locally to check the resulting link structure.

wilsonhj:
Ok. I kind of figured. I’ll have to do this again. How are you building
locally? Are you running Julia 1.x in Atom? Is there documentation for @ref
@id ?

ChrisRackauckas:
>How are you building locally? 

https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/master/docs/make.jl#L1-L112

>Is there documentation for @ref @id ?

https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@ref-link-1 and search around for id.

wilsonhj:
Chris, I want to do this correctly. Links that can be replaced with @ref
must be in the juliaDiffEq codebase? The headers must match exactly to be
replaced by @ref?

ChrisRackauckas:
>Chris, I want to do this correctly. Links that can be replaced with @ref must be in the juliaDiffEq codebase? The headers must match exactly to be replaced by @ref?

It must match the header or an id placed on it.

@kanav99 help explain if you can. But I really think

https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/309/files

is sufficient documentation of the idea.

kanav99:
Hey @wilsonhj The issue you are trying to fix is something we already are trying to fix (#309). You should refer to [Julia Documentation](https://docs.julialang.org/en/v1/manual/documentation/) page (for seeing how docstrings work and [Documenter.jl](https://juliadocs.github.io/Documenter.jl/stable/) to see how you deploy it to a webpage.
Broken links issue is already fixed by #309. So you can probably help us find and fix any further issues if you feel so.

ChrisRackauckas:
Instead of focusing on links, @wilsonhj how about grammar/typos for a page?

PR #315, opened 2019-12-17
extracting partials from sol_dual in `using ForwardDiff.jl`

hpodhaisky:
sol_dual contains the solution and the duals, hence the offset [i+1] is needed to extract the i-th derivative


ChrisRackauckas:
Yes, thanks!

PR #316, opened 2019-12-17
change remaining y(t) to u(t) for concistency

hpodhaisky:
- the i-th component was promised in the wording thus it should be given in the formula
- since solution is named u(t) everywhere else, the remaining occurrence of y(t) needed to be eliminated
- I am not sure whether the sentence with Jv with AD is clear now. I have added that this is without truncation error (but in the f(d) formula still needed the \approx because higher order terms are omitted).

### DIFF
--- a/docs/src/analysis/sensitivity.md
+++ b/docs/src/analysis/sensitivity.md
@@ -1,11 +1,11 @@
 # [Local Sensitivity Analysis (Automatic Differentiation)](@id sensitivity)
 
 Sensitivity analysis for ODE models is provided by the DiffEq suite. The model
-sensitivities are the derivatives of the solution with respect to the
+sensitivities are the derivatives of the solution `u(t)` with respect to the
 parameters. Specifically, the local sensitivity of the solution to a parameter
 is defined by how much the solution would change by changes in the parameter,
 i.e. the sensitivity of the ith independent variable to the jth parameter is
-`` \frac{\partial u}{\partial p_{j}}``.
+`` \frac{\partial u_i}{\partial p_{j}}``.
 
 Sensitivity analysis serves two major purposes. On one hand, the sensitivities
 are diagnostics of the model which are useful for understand how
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>I am not sure whether the sentence with Jv with AD is clear now. I have added that this is without truncation error (but in the f(d) formula still needed the \approx because higher order terms are omitted).

No, this is not approximate. This calculates precisely (with some floating point error of course) Jv. There are no higher terms that are dropped off. This is mentioned here: https://mitmath.github.io/18337/lecture9/autodiff_dimensions

hpodhaisky:
Jv is exact and this is what you want and therefore I added ("no truncation error") when the AD approach is mentioned.

The formula in which I changed to `\approx` contains f(d) on the left. This is the little o(eps) in your document "In the following we will ignore higher-order terms; formally we set ϵ^2=0". Your paragraph "Dual numbers" has the `approx` hidden in the wording "Note that the chain rule ... **to first order**." 

Maybe, "=" is better, but then the "to first order" added here wouldn't do much harm -- and mathematically it is needed.

ChrisRackauckas:
>formally we set ϵ^2=0

No, it truly is a nilpotent algebra. Dual number algebra is not a first order approximation, it's an implementation of smooth infinitesimal arithmetic which is an equivalent definition of the derivative.

hpodhaisky:
It depends on the definition of epsilon which is no part of this documentation. For 99% of the readers of your page it will be a small real number, perhaps sqrt(macheps), and they need O(eps^2), non-standard analysis is, non-standard.
I am happy if you leave "=" if you really want to. But renaming y(t) to u(t) is still helpful, I hope.

hpodhaisky:
To make sure that we are talking about the same:

The dual approach gives Jv exactly. No truncation. In an elegant way. And "no truncation" can be added.

The definition of the derivative is

f(x+eps v) = f(x) + eps Jv + o(eps^2)

and for eps -> 0 the change in direction is the Jacobian times v. Thus the formula
with f(d) needs an approx or a "to first order".

ChrisRackauckas:
https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/316/files#diff-4836821741b0b512fd723b0193066285R239

here epsilon is a dimensional signifier since it's explicit that `d` is a dual number, so the next line is exact.

hpodhaisky:
Ok. Please leave "=" there :-). Thanks for the explanation.

(And I hope that readers will think of epsilon being a "dimensional signifier" rather than a small positive number.)

PR #317, opened 2019-12-18
extract_local_sensitivities(sol,i::Integer) broken

hpodhaisky:
extract_local_sensitivities(sol,i::Integer) doesn't work the way it is described,  looking at the source it can be seen that i is not the component but the discrete time ith step -- which is not what a user is interested in (since the time points are chosen by the adaptive time stepping)

The correct implementation of  extract_local_sensitivities(sol,i::Integer) would have
sol[i, :] somewhere.

I think, it is enough to provide `extract_local_sensitivities(sol)` and ` extract_local_sensitivities(sol,t::T) where T <: AbstractFloat` and delete `extract_local_sensitivities(sol, i)` from documentation and code basis.

### DIFF
--- a/docs/src/analysis/sensitivity.md
+++ b/docs/src/analysis/sensitivity.md
@@ -301,14 +301,13 @@ We can use the following helper functions to extract the sensitivity information
 
 ```julia
 x,dp = extract_local_sensitivities(sol)
-x,dp = extract_local_sensitivities(sol,i)
 x,dp = extract_local_sensitivities(sol,t)
 ```
 
 In each case, `x` is the ODE values and `dp` is the matrix of sensitivities
 where `dp[i]` is the gradient of component `i` by the parameters. The first gives
-the full timeseries of values. The second returns the `i`th values, while the third
-interpolates to calculate the sensitivities at time `t`. For example, if we do:
+the full timeseries of values. The second interpolates to calculate the sensitivities at time `t`.
+For example, if we do:
 
 ```julia
 x,dp = extract_local_sensitivities(sol)


ChrisRackauckas:
>I think, it is enough to provide extract_local_sensitivities(sol) and extract_local_sensitivities(sol,t::T) where T <: AbstractFloat and delete extract_local_sensitivities(sol, i) from documentation and code basis.

The integer one is convenient if the person used adaptive time stepping without interpolants and wants to match up with those values.

ChrisRackauckas:
>extract_local_sensitivities(sol,i::Integer) broken

Do you have an MWE for this? I think it's still passing tests?

hpodhaisky:
"broken" is perhaps not the correct word. But " is the gradient of component i by the parameters " has a different meaning of "component i" as in the rest of the document. 
`sol(t, i)` computes the adjoint variable corresponding to all components of the solution evaluated a time t_i .

ChrisRackauckas:
Change it to j?

hpodhaisky:
just to add something like "component at time t_i" would resolve any potential misunderstanding here -- I am assuming that you don't want to edit too much, thus minimal incremental changes are best

ChrisRackauckas:
>just to add something like "component at time t_i" would resolve any potential misunderstanding here -- I am assuming that you don't want to edit too much, thus minimal incremental changes are best

Yeah that sounds good. It's more about not doing breaking changes if not necessary: thousands of people are using these tools and so deleting functions causes unnecessary havoc if it's doing its job. This has a job, but just poor documentation, so the thing to fix is just something like that which clarifies what `i` means in this case.

PR #318, opened 2019-12-18
introduction to adjoint sensitivity analysis

hpodhaisky:
I have tried to make the use of formulae more consistent. In particular, the arguments of the functions should be clear, using u(p), u(t), u(t,p) interchangeably is too confusing. There is still more work to do in this direction, however, I would like to see first whether you agree with this PR. The term "functional" should not be used for ``g(u,t,p)`` but for ``G(u,p)`` (not yet edited).

### DIFF
--- a/docs/src/analysis/sensitivity.md
+++ b/docs/src/analysis/sensitivity.md
@@ -378,26 +378,27 @@ will plot the expanded system).
 Adjoint sensitivity analysis is used to find the gradient of the solution
 with respect to some functional of the solution. In many cases this is used
 in an optimization problem to return the gradient with respect to some cost
-function. It is equivalent to "backpropogation" or reverse-mode automatic
+function. It is equivalent to "backpropagation" or reverse-mode automatic
 differentiation of a differential equation.
 
 ### Adjoint Sensitivity Analysis via adjoint_sensitivities
 
-This adjoint requires the definition of some scalar functional ``g(u,p,t)``
-where ``u`` is the (numerical) solution to the differential equation.
+This adjoint requires the definition of some scalar functional ``g(u,p)``
+where ``u(t,p)`` is the (numerical) solution to the differential equation
+``d/dt u(t,p)=f(t,u,p)`` with ``t\in [0,T]`` and ``u(t_0,p)=u_0``.
 Adjoint sensitivity analysis finds the gradient of
 
 ```math
-G(u,p)=G(u(p))=\int_{t_{0}}^{T}g(u(t,p))dt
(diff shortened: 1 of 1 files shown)


docs/src/analysis/sensitivity.md
```diff
@@ -378,26 +378,27 @@ will plot the expanded system).
...
-function. It is equivalent to "backpropogation" or reverse-mode automatic
+function. It is equivalent to "backpropagation" or reverse-mode automatic
 differentiation of a differential equation.
 
 ### Adjoint Sensitivity Analysis via adjoint_sensitivities
 
-This adjoint requires the definition of some scalar functional ``g(u,p,t)``
-where ``u`` is the (numerical) solution to the differential equation.
+This adjoint requires the definition of some scalar functional ``g(u,p)``
+where ``u(t,p)`` is the (numerical) solution to the differential equation
```
ChrisRackauckas: p,t is used everywhere in code, so keep the ordering consistent.

docs/src/analysis/sensitivity.md
```diff
@@ -378,26 +378,27 @@ will plot the expanded system).
...
+``d/dt u(t,p)=f(t,u,p)`` with ``t\in [0,T]`` and ``u(t_0,p)=u_0``.
```
ChrisRackauckas: u,p,t

docs/src/analysis/sensitivity.md
```diff
@@ -416,7 +417,7 @@ the adjoint sensitivity of some discontinuous functional of the solution. One
 canonical function is the L2 loss against some data points, that is:
 
 ```math
-L(u,p,t)=\sum_{i=1}^{n}\Vert\tilde{u}(t_{i})-u(t_{i},p)\Vert^{2}
```
ChrisRackauckas: it could be explicitly time dependent IIRC

ChrisRackauckas:
Yeah, the adjoint docs definitely need some work. I don't even think the energy functional one is fully correct.

docs/src/analysis/sensitivity.md
hpodhaisky: I am sorry, yes, please use an ordering with is consistent with the code, but use it consistently.

docs/src/analysis/sensitivity.md
hpodhaisky: But is does not, t is gone after the summation over time. The same was with G. You have the functional G(u,p) which operates on the function t->u(t). Both functionals are continuous as mappings from u -> R, thus "discontinous functional" is better rephrased (like "discrete functional", "functional evaluating a discrete points etc.).

PR #320, opened 2019-12-25
Restructure GSA documentation

ChrisRackauckas:
This also proposes a few changes to the interface:

- Everything should get the same `p_range` call
- Anything else should end up as keyword arguments in the methods
- Morris should somehow be reduced to an `N`, and then some extra (optional) details in the methods
- Sobol should move its error estimator argument
- Regression needs an interface change

### DIFF
--- a/docs/src/analysis/global_sensitivity.md
+++ b/docs/src/analysis/global_sensitivity.md
@@ -81,7 +81,7 @@ individual contribution.
 The `Sobol` object has as its fields the `order` of the indices to be estimated. 
 
 - `order` - the order of the indices to calculate. Defaults to `[0,1]`, which means the
-  Total and first order indices. Currently only these are allowed.
+  Total and First order indices. Passing `2` enables calculation of the Second order indices as well.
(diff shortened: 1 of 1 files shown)


docs/src/analysis/global_sensitivity.md
```diff
@@ -39,26 +76,16 @@ interaction and the contribution of the parameters individually and gives the
...
-  generated out of which `num_trajectory` matrices with the highest spread are used in calculation.
-  4. `len_design_mat` - The size of a design matrix.
-
-  `param_range` requires an array of 2-tuples with the lower bound
-and the upper bound.
+The `Sobol` object has as its fields the `order` of the indices to be estimated. 
 
-## Sobol Method
+- `order` - the order of the indices to calculate. Defaults to `[0,1]`, which means the
+  Total and first order indices. Currently only these are allowed.
```
Vaibhavdixit02:
```suggestion
  Total and First order indices. Passing `2` enables calculation of the Second order indices as well.
```

PR #322, opened 2020-01-10
Overhaul the sensitivity analysis page

ChrisRackauckas:
- Focus high level on `concrete_solve`
- Document the `SensitivityAlgorithm` choices
- All adjoints return u0 derivatives
- Put the lower level interfaces below, and explain when they might be useful


ChrisRackauckas:
At least merging for now since it's a whole lot better than before. It should all be good?

PR #328, opened 2020-02-17
Mass matrix example: fix index

scheidan:
also rearrange the example code to match the equation better


ChrisRackauckas:
Thanks!

PR #338, opened 2020-03-16
Added docs for uprev2

kamalojasv181:


ChrisRackauckas:
This is not a user-facing detail. If it goes anywhere, it would be in devdocs.

PR #341, opened 2020-03-20
update old function calls

ric-cioffi:


ChrisRackauckas:
awesome thanks. Wow those were old :P

PR #345, opened 2020-04-01
Fix reduction func in ensembles example #3

jlumpe:
Fixed a redundant `$condition ? true : false` ternary statement and used the extra space to add spaces around binary operators.


ChrisRackauckas:
oh yeah oops. Thanks!

PR #357, opened 2020-05-10
Update RegularJumps documentation for the matrix-free style

ChrisRackauckas:
@TorkelE @isaacsas this was updated awhile ago, but never got upstreamed into DiffEqBio, but now is the time to probably handle it. (This is what throws the warning: it has a fallback so DiffEqBio still works, just not optimally)

### DIFF
--- a/docs/src/types/jump_types.md
+++ b/docs/src/types/jump_types.md
@@ -59,20 +59,16 @@ while `RegularJump`s require special algorithms.
 The constructor for a `RegularJump` is:
 
 ```julia
-RegularJump(rate,c,c_prototype;mark_dist = nothing,constant_c = false)
+RegularJump(rate,c,numjumps;mark_dist = nothing)
 ```
 
 - `rate(out,u,p,t)` is the function which computes the rate for every regular
   jump process
-- `c(dc,u,p,t,mark)` is the current Stoichiometry matrix for each jump process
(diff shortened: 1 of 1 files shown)


isaacsas:
We can make sure to use the updated version in the new DEBio version. In MT, does it make sense to create a separate `RegularJumpSystem` instead of trying to mix them with the other jump types? We could always add a conversion function between the two I'd think.

ChrisRackauckas:
Yeah, it would make sense to make a `RegularJumpSystem` in DiffEqBio and conversions to it which generates the whole system as regular jumps. This will then give us a playground to start working on more refined conversions, like `MixedRegularSSA` that has both versions of the jumps and sets up the parameters in a nice way to allow switching between the two when `x<switch`, and possibly make this as a whole ODEProblem with a second switch, etc. Which is then incentive to get the tau-leaping into StochasticDiffEq and start doing all sorts of cool things... Basically `RegularJumpSystem` will be a great stepping stone, so I'd love to have it.

isaacsas:
Yeah, heading towards hybrid system representations was exactly my thought too.

isaacsas:
Once we get jumpsystems, dep graphs, and conversions from reaction systems for jumps, I’ll add reg jump systems (if no one else has looked at it by that point). That will give parity with the current DEBio functionality I think...

PR #372, opened 2020-06-16
Fix Typo in "solve Differentiation Examples"

Janssena:
Using sum_of_solution(u0,p) to calculate the gradient results in (nothing, nothing) because the non-remade prob was referenced in the solve call.


Janssena:
Push to netlify failed with: `{u'message': u'Access Denied: Missing access token', u'code': 401}`. How do I fix this?

ChrisRackauckas:
thanks!

Janssena:
Thanks for merging!

PR #381, opened 2020-07-10
Swapped roles of parameter 1 and 2 in 'Generalized Likelihood Example'

SteffenPL:
The text in https://docs.sciml.ai/stable/analysis/parameter_estimation/#Generalized-Likelihood-Example-1 does not match with the example. It should be like this: parameter 1 is clearly 1.5 and we need to find parameter 2 using BlackBoxOptim.jl.

In the current text, parameter 2 is said to be 1.5 and the influence of parameter 1 on the objective is plotted.
I changed the plots such that the more interesting dependency of the objective w.r.t. parameter 2 is shown.

I also renamed `range` to `prange` to avoid conflict with `Base.range`. (But this is not important at all.)


ChrisRackauckas:
hehe oops! Thanks!

PR #390, opened 2020-07-30
Reflect support of callbacks for SSAStepper

manuel-rhdt:
This functionality was introduced in https://github.com/SciML/DiffEqJump.jl/pull/120


ChrisRackauckas:
Thanks!

PR #391, opened 2020-08-02
Added notes for solution statistics `sol.destats`

cuihantao:
Added docs for `sol.destats` for viewing internal statistics to the "special fields" section.


ChrisRackauckas:
Thanks!

PR #394, opened 2020-08-09
update Docs for Catalyst

isaacsas:


isaacsas:
Not sure the issue here; it builds fine locally.

isaacsas:
Only thing I've been able to resolve is that the issue is not present on newer Documenter.jl versions (i.e. latest v.25 releases).

ChrisRackauckas:
I don't think we need the using Catalyst in these docs. I think that might just be a relic.

ChrisRackauckas:
The issue seems to just be netlify. Fingers crossed.

isaacsas:
It may be a general v.24 issue; I get a local failure with the same error on v24.11. Unfortunately the error message is not helpful to what the issue is.

PR #399, opened 2020-08-14
Update advanced_ode_example.md

hwpang:
This is a very minor change. This example inputs parameter as a Tuple, which caused error at the `size(prob.p)` used at src/systems/diffeqs/modelingtoolkitize.jl line 11. Inputting as Array avoids this problem.


ChrisRackauckas:
Great thanks!

PR #407, opened 2020-09-01
Fix typo in DiffEq Operator docs

briochemc:
There is no `DiffEqLinearOperator` AFAIK, so I'm assuming it was meant to say `DiffEqArrayOperator` instead (although I'm actually not sure I'm correct).


ChrisRackauckas:
No, it means the abstract type.

briochemc:
> No, it means the abstract type.

OK, made a new commit! 🙂

ChrisRackauckas:
Thanks!

PR #412, opened 2020-09-22
Resolve an inconsistency in the pendulum example

rozsasarpi:
The Julia code and the supporting text were not consistent, the text was updated to match the Julia code.


ChrisRackauckas:
Thanks!

PR #414, opened 2020-10-08
Fix typo and attribution for RKO65

Gyoshi:
The in-code documentation should also be updated with the article, because the only way I could find the source of the `RKO65` method was finding https://github.com/SciML/OrdinaryDiffEq.jl/issues/864 via git blame. I don't know where to do this though.


ChrisRackauckas:
>I don't know where to do this though.

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/algorithms.jl#L532

PR #418, opened 2020-11-06
Update parameter_estimation.md

mkg33:
Added an introduction containing the overviews of packages relevant to parameter estimation.


ChrisRackauckas:
Great! I think when the other docs are up and running we'd delete the portions below this as well.

mkg33:
That's right, I've already marked it for future deletion in my notes.

PR #425, opened 2020-12-10
Add missing footnote and fix duplicated heading

### LINKED ISSUE #424: Missing footnote on non-autonomous solvers page
The footnote associated with the `iop` option of the exponential method is missing.

imciner2:
This PR contains two minor fixes to the docs:
1. Add the missing footnote for the IOP reference in the `LinearExponential` integrator documentation.
2. Change the section label for the non-autonomous problems section to not be the same as the label for the ODE problems section. The non-autonomous section is actually never referenced, so no places need to be updated with the change (but the ODE section is referenced, so this caused confusion over which section to link to).

Fixes #424


ChrisRackauckas:
Thanks!

PR #453, opened 2021-03-22
fix calling signature of g for ManifoldProjection

ranocha:
The documentation seems to be outdated - the signature did not match the convention used in the source code.


ChrisRackauckas:
Thanks!

PR #454, opened 2021-04-03
Fix order of return values in integrator intervals() example

Leixb:
Small error on the order of the return values for `intervals(integrator)` example 

It should be: `uprev, tprev, u, t` ([integrator_interface.jl#L439](https://github.com/SciML/SciMLBase.jl/blob/ca50226c8db0f446fd28e70344f34a86c962aade/src/integrator_interface.jl#L439))


ChrisRackauckas:
Thanks!

PR #459, opened 2021-04-24
Fixes to make adjoint_sensitivity example run

olof3:
Also two minor changes to the math. I hope it looks alright.


ChrisRackauckas (approved):
Thanks!

PR #476, opened 2021-07-02
Add diffusion-implicit heat equation tutorial

charleskawczynski:
This PR adds a diffusion-implicit heat tutorial. cc @ChrisRackauckas


docs/src/tutorials/diffusion_implicit_heat_equation.md
```diff
@@ -0,0 +1,186 @@
...
+Δt = 1000;                      # timestep size
+N_t = 10;                       # number of timesteps to take
+FT = Float64;                   # float type
+Δz = FT(b-a)/FT(n)
+Δz² = Δz^2;
+∇²_op = [1/Δz², -2/Δz², 1/Δz²]; # interior Laplacian operator
+∇T_bottom = 10;                 # Temperature gradient at the top
+T_top = 1;                      # Temperature at the bottom
+S(z) = β*sin(γ*z)               # source term, (sin for easy integration)
+zf = range(a, b, length=n+1);   # coordinates on cell faces
```
ChrisRackauckas, line 36: Is this actually needed for the tutorial? It seems like this is doing a lot of unnecessary things. Seems more like a https://github.com/SciML/SciMLTutorials.jl tutorial instead of something streamlined for the docs.

docs/src/tutorials/diffusion_implicit_heat_equation.md
```diff
@@ -0,0 +1,186 @@
+# Solving the heat equation with diffusion-implicit time-stepping.
```
ChrisRackauckas, line 1:
```suggestion
# IMEX Integrators
```

This title is more informative as to what the tutorial is actually trying to teach.

docs/src/tutorials/diffusion_implicit_heat_equation.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+and Neumann boundary stencil & source:
+```math
+∇T_{bottom} n̂ = (T[g] - T[i])/(2Δz), \qquad    n̂ = [-1,1] ∈ [zmin,zmax] \\
+T[i] + 2 Δz ∇T_{bottom} n̂ = T[g] \\
+∂_t T = α (((T[i] + 2 Δz ∇T_{bottom} n̂) - T[b])/Δz - (T[b] - T[i])/Δz)/Δz + S \\
+∂_t T = α (((T[i]) - T[b])/Δz - (T[b] - T[i])/Δz)/Δz + S + α 2 Δz ∇T_{bottom} n̂/Δz² \\
+∂_t T = α (2 T[i] - 2 T[b])/Δz² + S + 2α ∇T_{bottom} n̂/Δz
+```
```
ChrisRackauckas, line 95: Yeah, all of this is too much build up and loses the message. I think this should all go to https://github.com/SciML/SciMLTutorials.jl . It would be good to get an IMEX in here, but such a tutorial would want to have no more than a 1 paragraph problem statement, which is very different from what this is trying to do. This is trying to show a workflow specific to a use cases, which is what the extended tutorial set is made for.

docs/src/tutorials/diffusion_implicit_heat_equation.md
```diff
@@ -0,0 +1,186 @@
...
+prob = SplitODEProblem(
+    SciMLBase.DiffEqArrayOperator(
+        D,
+    ),
+    rhs!,
+    T,
+    tspan,
+    params
+)
+alg = IMEXEuler(linsolve=LinSolveFactorize(lu!))
```
ChrisRackauckas, line 159: Are you sure IMEXEuler is a good idea here? I would assume KenCarp4 etc. would perform better.

docs/src/tutorials/diffusion_implicit_heat_equation.md
charleskawczynski: Ah, yeah, I wasn’t exactly sure where to put this. I’m fine with moving it (the whole tutorial, right?) there

docs/src/tutorials/diffusion_implicit_heat_equation.md
ChrisRackauckas: If it's moved, the original title makes sense though.

docs/src/tutorials/diffusion_implicit_heat_equation.md
ChrisRackauckas: Yeah, let's go with the move. I am putting up a bounty soon anyways to try and find a way to incorporate those tutorials and the benchmarks into the documentation. It's mostly split so that it's easy to regenerate the results using the AMDCI computers, allowing you to go ham and make it a big equation and really demonstrate sparsity handling etc. I hope that it will soon then feed back nicely into the main docs.

docs/src/tutorials/diffusion_implicit_heat_equation.md
charleskawczynski: Yeah, probably, I was just trying to keep certain things simple.

charleskawczynski:
Closing in favor of [SciMLTutorials#416](https://github.com/SciML/SciMLTutorials.jl/pull/416)

PR #481, opened 2021-07-15
IDA remove prec_side option

sjdaines:
IDA supports left preconditioning only.
https://github.com/SciML/Sundials.jl/issues/317
https://github.com/SciML/Sundials.jl/pull/319


ChrisRackauckas:
Thanks!

PR #484, opened 2021-07-22
Fix broken link

lamorton:
Turing evidently renamed their tutorial pages.


ChrisRackauckas:
I think a link in the DiffEqFlux docs would be broken too then?

PR #494, opened 2021-08-18
update the DiffEqJump tutorial

isaacsas:
This is pretty big rewrite of the DiffEqJump / Gillespie tutorial. Some key changes:

1. Added a section showing the mathematical jump process equations (for users who aren't familiar with reactions).
2. Rewrote the bio modeling section some.
3. Flow is now: Mathematical Model -> Biological Model/Explanation for less Mathematical Users -> Modeling with Catalyst -> Modeling using the low-level DiffEqJump Interface -> Jump Types -> ...
4. Added a FAQ that collates some common questions with answers like remaking and using discrete callbacks.
5. Switched MassActionJumps to now use parameter vectors/tuples.
6. Tweaked the VariableRateJump example rate/affect to keep the system non-negative (at least in the absence of continuous noise).


docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -1,135 +1,297 @@
-# Discrete Stochastic (Gillespie) Equations
+# Continuous-Time Jump Processes
```
ChrisRackauckas, line 1: I feel like we need to put Gillespie in there just for the search results. Continuous-Time Jump Processes (Gillespie) or something.

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -443,8 +577,64 @@ One such algorithm is `SimpleTauLeaping`, which we use as follows:
...
+Random number generators can be passed to `JumpProblem` via the `rng` keyword
+argument. Continuing the previous example:
+
+```julia
+#] add RandomNumbers
+using RandomNumbers
+jprob = JumpProblem(dprob, Direct(), maj, rng=Xorshifts.Xoroshiro128Star(rand(UInt64)))
+```
+uses the `Xoroshiro128Star` generator from
+[RandomNumbers.jl](https://github.com/JuliaRandom/RandomNumbers.jl).
```
ChrisRackauckas, line 640: This may even make sense to repeat this in the Catalyst docs.

ChrisRackauckas:
This is a really fantastic update. It would probably make sense to similarly see if an update to the jump diffusion is required, and maybe mention some of the StochasticDiffEq methods?

docs/src/tutorials/discrete_stochastic_example.md
isaacsas: Yeah I'll update that. I noticed a few other tweaks on the JumpProblem page to make too (I only want to expose the new MassActionJump interface in the docs).

docs/src/tutorials/discrete_stochastic_example.md
isaacsas: Or we can at least link to it from there.

ChrisRackauckas: that's a better idea.

ChrisRackauckas (approved):

isaacsas:
@ChrisRackauckas this should now be good to go. Updating the jump diffusion tutorial some and switching to the StochasticDiffEq tau-leaping as the recommended regular jump approach makes sense. Let's do that as a subsequent doc update since they will also be some work to write up.

I'll try to also followup with Catalyst doc tweaks and some links in the DiffEqJump README once this goes live.

PR #497, opened 2021-08-31
Typo in common_solver_opts

nbelakovski:
estimatoes sounds like a delicious variety of tomatoes, but unfortunately for us all the correct spelling is estimates


ChrisRackauckas:
Ehh, less tasty.

PR #502, opened 2021-09-16
add strange attractor odes

anandijain:
will this just work or do i need to change a compat somewhere


ChrisRackauckas:
That should work.

anandijain:
does it need a diffeq version bump to deploy? if so lets i'd opt to wait for other stuff rather than patch for this

ChrisRackauckas:
it'll be on dev.

anandijain:
https://github.com/SciML/DiffEqDocs.jl/runs/3625228913#step:5:16 https://diffeq.sciml.ai/dev/types/ode_types/ 
hmm seems like its missing the docstrings, but i don't see anything obviously wrong

PR #506, opened 2021-10-15
Updated the overal description of callback types

hurak:
Highlighted that the difference between `ContinuousCallback` and `DiscreteCallback` is that the former checks hitting-the-zero even **within the integration step**, while the latter only checks if the condition is true **at the end of (every) integration step**.

### DIFF
--- a/docs/src/features/callback_functions.md
+++ b/docs/src/features/callback_functions.md
@@ -7,12 +7,10 @@ types can be used to build libraries of extension behavior.
 
 ## The Callback Types
 
-The callback types are defined as follows. There are three primitive callback types: the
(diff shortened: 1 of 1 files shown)


hurak:
Well, I am still hesitating if perhaps a more change in the terminology should be introduced: while in both cases we have `condition` functions, in the `ContinuousCallback` case it is really just a function that returns reals and the condition-triggering event is only given by `condition == 0`, and in the `DiscreteCallback` case the `condition` is directly the *condition* itself (hence it evaluates to boolean). How about renaming the `condition` in the docs for the former to something like `condition_fcn` or `event_fcn`?

ChrisRackauckas:
Yeah, if that helps people it could make sense to do the name change in the docs. No reason to delay this though.

Issue #509, opened 2021-10-28
Documentation: expression for SDE with jumps


tom-plaa:
On the Jump Diffusions tutorial at
DiffEqDocs.jl/docs/src/tutorials/jump_diffusion.md

EDIT: Also seems to be the same in docs/src/types/jump_types.md

I find it confusing to read
```math
\frac{du}{dt} = f(u,p,t) + Σgᵢ(u,t)dWⁱ + Σ h_i(u,p,t)N_i(t)
```

Because it implies \int{ g_i(u,t) dW_i(t) dt} on the diffusion term
And \int{ h_i(u,t) N(t) dt} on the jump term.

I would suggest to adopt the notation used in the Itô diffusions tutorial, something akin to:

```math
du = f(u,p,t)dt + Σgᵢ(u,p,t)dWⁱ(t) + Σ h_i(u,p,t)dNⁱ(t)
```
Because now it implies the expected integrals \int{ g_i(u,p,t) dW_i(t)} and \int{ h_i(u,p,t) dN_i(t)}

By the way, I am assuming that these are the types on integrals being defined by the Problem interface for Jumps, I may be wrong (if so, can you please clarify how they are being defined?).

I assume that having du/dt on the left might have been influenced by the ODE methodology as trying to make it easier to associate with the actual code that one has to write for the integrand functions, but I think this merit might not be worth the confusion that arises regarding how the module is integrating with respect to jump processes.

Thanks for this module, very impressive work!

ChrisRackauckas:
Yes, your proposed changes would be an improvement. Please submit a PR. Thanks!

tom-plaa:
I will do so as soon as I can, thanks!

PR #511, opened 2021-11-03
Rewrite mathematical expression for differential equations with jump …

tom-plaa:
Rewrite the mathematical expressions for equations with jump components, so that they imply an integral with respect to the Poisson counting measure.
#509


ChrisRackauckas:
Thanks!

Issue #513, opened 2021-12-03
Link to developer documentation in Overview is broken


youainti:
While reading through the documentation [here (lines 89 and 94)](https://github.com/SciML/DiffEqDocs.jl/blob/master/docs/src/basics/overview.md), I found that the link to the developer documentation is broken. I've tried to locate the appropriate website but have been unsucessful.

I just wanted to raise the issue so that someone with a little more information as to where such documentation may have moved to can fix it, or point me in the right direction so that I can fix it.

youainti:
I just found the new link [https://devdocs.sciml.ai/dev/](https://devdocs.sciml.ai/dev/). Are there any contribution details I need to be aware of before submitting corrections?

ChrisRackauckas:
Nope, it's all MIT licensed etc. so a PR would be perfect. Thanks for finding this.

PR #514, opened 2021-12-04
Corrected prominent links to developer documentation in overview

youainti:
Pretty straightforward, fixed two links. I've identified others, but I'm leaving that for another day/issue. As background, this was referenced in #513 .


ChrisRackauckas:
Thanks! If you could point out the other broken links you found that would be great. I can go clean them up, I just don't know where they are.

youainti:
I'm thinking of making a script that will pull all the links and then test them for 404 errors. If I find some time over the next few weeks I'll do that.

PR #515, opened 2021-12-06
Update discrete_types.md

hstrey:
Corrected the function map definition to reflect the actual implementation.  For scale_by_time=true, I replaced t_{n} with t_{n+1} since this is how the solver is implemented.  In addition, I replaced du, u, t with unp1, un, tnp1 to clarify how f implements the map.

### DIFF
--- a/docs/src/types/discrete_types.md
+++ b/docs/src/types/discrete_types.md
@@ -6,10 +6,10 @@ To define an Discrete Problem, you simply need to give the function ``f`` and th
 condition ``u₀`` which define a function map:
 
 ```math
-u_{n+1} = f(u,p,t_{n+1})
+u_{n+1} = f(u_{n},p,t_{n+1})
 ```
 
-`f` should be specified as `f(u,p,t)` (or in-place as `f(du,u,p,t)`), and `u₀` should
+`f` should be specified as `f(un,p,tnp1)` (or in-place as `f(unp1,un,p,tnp1)`), and `u₀` should
 be an AbstractArray (or number) whose geometry matches the desired geometry of `u`.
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Seems like it would be good to make it `t_n`?

hstrey:
you mean to implement the solver to use t_n?  If you look at the solver implementation and the solver documentation it uses t_{n+1}

hstrey:
it makes sense to have u_n+1 only depend on the past (u_n,t_n).  That is more Markovian.

ChrisRackauckas:
Yeah change the docs to that and let's fix the package. That makes more sense.

hstrey:
I agree.  It would make sense to change all the behavior (scale_by_time = true or false) to behave the same u_n+1 = f(u_n,p,t_n).  Correct?  I can make the changes to the documentation and then also update the discrete solver docs.  @ChrisRackauckas has to change the code.   But then we need to rewrite the discrete test suite.

hstrey:
ok, changed the docs to t_n.  Who uses the discrete solvers?  I am worried that it may break someone's code if we change the behavior.  It's Gillespie using it?

ChrisRackauckas:
Gillespie uses it but by definition that cannot be `t`-dependent, and that was the main reason for those methods in the first place which is why that part of the code is not as well-tested. DynamicalSystems.jl uses it widely but for autonomous dynamical systems, so again, no `t`. 😅

hstrey:
ok, I also changed the documentation for the discrete solver.  If you are happy, you can merge with the master.  If you change the behavior of the solver, then I can help update the tests.

PR #516, opened 2021-12-14
correct error in integrator interface use

codercahol:
integrator interface has changed since this was last written


ChrisRackauckas:
awesome thanks.

PR #521, opened 2022-01-03
Explanation sensealg with ForwardDiff

ArnoStrouwen:
For users it might not be obvious enough from the current note how ForwardDiff is handled differently from other AD libraries.

### DIFF
--- a/docs/src/analysis/sensitivity.md
+++ b/docs/src/analysis/sensitivity.md
@@ -66,7 +66,7 @@ is used, i.e. going back to the AD mechanism.
     The behavior of ForwardDiff.jl is different from the other automatic differentiation libraries mentioned above.
     The `sensealg` keyword is ignored. Instead, the differential equations are solved using `Dual` numbers for `u0` and `p`.
     If only `p` is perturbed in the sensitivity analysis, but not `u0`, the state is still implemented as a `Dual` number.
-    ForwardDiff.jl is thus not compatible with a reverse pass through, nor with continuous forward sensitivity analysis.
+    ForwardDiff.jl will thus not dispatch into continuous forward nor adjoint sensitivity analysis even if a `sensealg` is provided.
 
 ### solve Differentiation Examples
 
--- a/docs/src/analysis/sensitivity.md
+++ b/docs/src/analysis/sensitivity.md
@@ -63,7 +63,10 @@ is used, i.e. going back to the AD mechanism.
 
 !!! note
 
-    ForwardDiff.jl only does forward differentiation pass through.
+    The behavior of ForwardDiff.jl is different from the other automatic differentiation libraries mentioned above.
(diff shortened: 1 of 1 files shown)


docs/src/analysis/sensitivity.md
```diff
@@ -63,7 +63,10 @@ is used, i.e. going back to the AD mechanism.
 
 !!! note
 
-    ForwardDiff.jl only does forward differentiation pass through.
+    The behavior of ForwardDiff.jl is different from the other automatic differentiation libraries mentioned above.
+    The `sensealg` keyword is ignored. Instead, the differential equations are solved using `Dual` numbers for `u0` and `p`.
+    If only `p` is perturbed in the sensitivity analysis, but not `u0`, the state is still implemented as a `Dual` number.
+    ForwardDiff.jl is thus not compatible with a reverse pass through, nor with continuous forward sensitivity analysis.
```
ChrisRackauckas, line 69: What do you mean by this line?

ArnoStrouwen: Which line do you mean? You quoted all of them.

I do not find it obvious from the docs that ForwardDiff does not work through sensealg like the other AD libraries do.
Why is it that Zygote can be combined with ForwardSensitivity() and ForwardDiffSensitivity(), but ForwardDiff not with TrackerAdjoint() or QuadratureAdjoint()? This should be possible with appropriate DiffRules?

I think a note needs to be added why it does not work like that. The current note is a bit vague.

ChrisRackauckas: "ForwardDiff.jl is thus not compatible with a reverse pass through, nor with continuous forward sensitivity analysis."

ChrisRackauckas: Oh I see.

ChrisRackauckas, line 69:
```suggestion
    ForwardDiff.jl will thus not dispatch into continuous forward nor adjoint sensitivity analysis even if a `sensealg` is provided.
```

ArnoStrouwen: The other AD libraries are able to do cross country AD, such as Zygote + ForwardDiffSensitivity(), and have the choice between discretize then optimize versus optimize then discretize. ForwardDiff does not have those options.

I guess the sentence could be confused with forward over reverse.

ChrisRackauckas:
> Why is it that Zygote can be combined with ForwardSensitivity() and ForwardDiffSensitivity(), but ForwardDiff not with TrackerAdjoint() or QuadratureAdjoint()? This should be possible with appropriate DiffRules?

DiffRules.jl only defines scalar rules. Instead you'd have to define a dual dispatch for arrays, like is done in Quadrature.jl https://github.com/SciML/Quadrature.jl/blob/master/src/Quadrature.jl#L583

PR #527, opened 2022-01-14
Fix type solve

ArnoStrouwen:


ChrisRackauckas:
Awesome thanks. Yeah my keyboard was broken and dropping some letters in the lower part of the keyboard 😆

PR #528, opened 2022-01-18
fix link to advanced ode example

vpuri3:
was broken


ChrisRackauckas:
Note the commit. Implicit refs are preferred over URLs.

PR #529, opened 2022-01-25
Clarify ways of specifying `f`

JordiBolibar:
Small modification to make it clearer for newcomers on the two main ways of specifying `f`. This is particularly relevant when working with non mutating AD packages such as Zygote. Until now, it was not obvious how to specify `f` if we cannot mutate an array.

Feel free to add any corrections on how to best integrate it to flow better.


ChrisRackauckas:
Thanks!

PR #536, opened 2022-02-18
update docs for stochastic iterated integrals

frankschae:
Should be merged after https://github.com/SciML/StochasticDiffEq.jl/pull/459.

@fkastner It would be great if you could quickly review this as well  :)

### DIFF
--- a/docs/src/solvers/sde_solve.md
+++ b/docs/src/solvers/sde_solve.md
@@ -89,7 +89,11 @@ approximation. The choices are:
   The package supports the schemes: `Fourier()`, `Milstein()`, `Wiktorsson()`,`MronRoe()`.
   The optimal algorithm is automatically selected based on the dimension of the Brownian
   process and the step size. By passing a specific scheme, e.g. `ii_approx=Fourier()`
-  methods can be manually selected.
+  methods can be manually selected. One must be careful when using the Levy area
+  approximations in conjunction with adaptivity (`adaptive=true`) because the Levy area
+  approximations draw random numbers that do not reflect the random numbers taken in a
+  previous rejected step. This leads to a bias that increases with an increasing number
+  of rejected steps.
 
 Example: `RKMilGeneral(;ii_approx=IILevyArea())`.
 
--- a/docs/src/solvers/sde_solve.md
+++ b/docs/src/solvers/sde_solve.md
@@ -8,9 +8,9 @@ do well. If the problem has additive noise, then `SOSRA` will be the
 optimal algorithm. At low tolerances (`<1e-4`?) `SRA3` will be more efficient,
 though `SOSRA` is more robust to stiffness. For commutative noise, `RKMilCommute`
 is a strong order 1.0 method which utilizes the commutivity property to greatly
-speed up the Wiktorsson approximation and can choose between Ito and Stratonovich.
-For non-commutative noise, difficult problems usually require adaptive time
-stepping in order to be efficient. In this case, `LambaEM` and `LambaEulerHeun`
+speed up the stochastic iterated integral approximation and can choose between Ito
+and Stratonovich. For non-commutative noise, difficult problems usually require adaptive
+time stepping in order to be efficient. In this case, `LambaEM` and `LambaEulerHeun`
 are adaptive and handle general non-diagonal problems (for Ito and Stratonovich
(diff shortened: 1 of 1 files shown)


fkastner:
What do you think about this:


## Iterated integral approximations
The difficulty of higher strong order integrators stems from the presence of iterated stochastic integrals
```math
I(h) = \int_0^h\int_0^sdW^1_tdW^2_s
```
in these schemes.

The approximation of these iterated integrals can be avoided, if the diffusion matrix satisfies the special commutativity condition given [above](@ref special noise forms).
Because of this, many methods are only applicable to problems that satisfy the commutativity condition.

However, the methods for general SDEs, like `RKMilGeneral`, perform a direct
approximation of the iterated integrals. For those methods, the algorithms have
an `ii_approx` keyword argument which allows to specify the method for the
approximation. The choices are:
- `IICommutative`: a simplification of the integral which assumes the noise commutativity
  property. If used on a non-commutative noise problem this will limit the strong convergence
  to 0.5.
- `IILevyArea`: computes the iterated integrals based on an approximation of the Levy area
  using the [LevyArea.jl](https://github.com/stochastics-uni-luebeck/LevyArea.jl) package:
  Kastner, F. and Rößler, A., [arXiv: 2201.08424](https://arxiv.org/abs/2201.08424)
  Kastner, F. and Rößler, A., LevyArea.jl, [10.5281/ZENODO.5883748](https://zenodo.org/record/5883749#.Yg-d698xmu4).
  The package supports the schemes: `Fourier()`, `Milstein()`, `Wiktorsson()`,`MronRoe()`.
  The optimal algorithm is automatically selected based on the dimension of the Brownian
  process and the step size. By passing a specific scheme, e.g. `ii_approx=Fourier()`
  methods can be manually selected.

Example: `RKMilGeneral(;ii_approx=IILevyArea())`.

fkastner:
Maybe also mention that it doesn't play nice with adaptivity?
Actually, what does it do if I try to use the adaptive methods with the Levy areas? Does it error?

frankschae:
It computes an error estimate as for the corresponding commutative noise case:
`RKMilGeneral()`:
https://github.com/SciML/StochasticDiffEq.jl/blob/c802f6ec6a8cb06e0d5797fc7605d7cb85fb9e25/src/perform_step/low_order.jl#L493
`RKMilCommute()`:
https://github.com/SciML/StochasticDiffEq.jl/blob/c802f6ec6a8cb06e0d5797fc7605d7cb85fb9e25/src/perform_step/low_order.jl#L368

I need to check again what this means in practice. Could it be that it keeps decreasing the step size and then throws an error? @ChrisRackauckas

ChrisRackauckas:
The issue is that if a step is rejected you won't be guaranteed something that "keeps" the values at `t + dt`, so it will likely be biased in some way. The commutative noise cases have extra stuff to handle this, but the Levy area approximations take random numbers that do no respect the random numbers taken in a previous rejected step (and it's not trivial to see how that would be done). It's hard to know how much bias it really introduces in practice, it depends on how aggressive the PI controller is and how many step rejections you're getting, so it's worth documenting as something that could potentially occur.

frankschae:
Did some short tests. At least, easy examples like:
```julia

using StochasticDiffEq, Random, Test

u₀ = [0.0,0.0,0.0]
f(u,p,t) = [0.0,0.0,0.0]
g(u,p,t) = [1.0 0.0;0.0 1.0;0.0 u[1]]
dt = 1//2^8
tspan = (0.0,1.0)
prob = SDEProblem(f,g,u₀,(0.0,1.0); noise_rate_prototype=zeros(3,2))

sol = solve(prob, RKMilGeneral(;ii_approx=IICommutative()), dt=dt, adaptive=true)
@test sol.retcode == :Success
@show sol.t
sol1 = solve(prob, RKMilGeneral(;ii_approx=IILevyArea()), dt=dt, adaptive=true)
@test sol1.retcode == :Success
@show sol1.t == sol.t  # True
```
result in the same stepping behavior.

ChrisRackauckas:
Yeah, the standard PI in that repo is tuned so that rejections are minimized. We know for example that RSwM2 is an approximation that isn't correct in theory, but in practice we haven't found a test that could solidly detect that 😅.

PR #544, opened 2022-04-06
Shadowing update

frankschae:
A bit more elaborate text regarding shadowing methods/chaos.

@ChrisRackauckas 
Do you see what's wrong with the section header? (Can we not use the bracket?)

https://diffeq.sciml.ai/stable/analysis/sensitivity/#Sensitivity-analysis-for-chaotic-systems-(shadowing-methods)(@id-shadowing_methods)


ChrisRackauckas:
Is there an issue with the section header? I don't see what would be an issue here?

ChrisRackauckas:
Maybe it can't be id'd?

frankschae:
So should we remove the id for/refs to the section?

PR #546, opened 2022-04-07
Rerun starts from one in EnsembleProblem

sosiristseng:
I also changed the order of `output_func` and `prob_func` for consistency with `EnsembleProblem`.


ChrisRackauckas:
Thanks!

PR #561, opened 2022-06-07
EnsembleProblem safetycopy clarification

Krastanov:
These are minor documentation clarification concerning Ensemble problems. If an SDE problem contains a custom noise process, `safetycopy` is insufficient for properly setting up multiple trajectories. The previous version of the documentation was not clear about that, leading to confusion like this one: https://discourse.julialang.org/t/sde-ensembleproblem-unhappy-with-trajectories-2/82098

### DIFF
--- a/docs/src/tutorials/sde_example.md
+++ b/docs/src/tutorials/sde_example.md
@@ -122,6 +122,9 @@ are added via `addprocs()`, but we can change this to use multithreading via
 sol = solve(ensembleprob,EnsembleThreads(),trajectories=1000)
 ```
 
+If you use a custom noise process, you might need to specify it in a custom `prob_func`
+in the `EnsembleProblem` constructor, as each trajectory needs its own noise process.
+
 Many more controls are defined at the [Ensemble simulations page](@ref ensemble), 
 including analysis tools.
 A very simple analysis can be done with the `EnsembleSummary`, which builds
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/sde_example.md
```diff
@@ -122,6 +122,9 @@ are added via `addprocs()`, but we can change this to use multithreading via
 sol = solve(ensembleprob,EnsembleThreads(),trajectories=1000)
 ```
 
+If you use a custom noise process, you might need to specify it in a custom `prob_func`
+in the `EnsembleProblem` constructor, as each trajectory needs its own noise process.
+
```
ChrisRackauckas, line 127:
```suggestion
```

!!! warn
    If you use a custom noise process, you might need to specify it in a custom `prob_func`
    in the `EnsembleProblem` constructor, as each trajectory needs its own noise process.

```

Krastanov:
@ChrisRackauckas , I think there might be a typo in your edit (the commit you have added to this PR). It looks like you deleted a closing triple backquote and now the code block does not end where it should.

ChrisRackauckas:
oh shit. yeah...

ChrisRackauckas:
https://github.com/SciML/DiffEqDocs.jl/commit/1c7f0c3ad5d0e11e31b292a57b0bb49123ca7cd7

PR #562, opened 2022-06-10
Update ensemble.md

chriselrod:
Before
```julia
julia> @time solve(ensemble_prob,SRIW1(),trajectories=10);
  6.432733 seconds (16.81 M allocations: 1.172 GiB, 6.05% gc time, 99.99% compilation time)

julia> @time solve(ensemble_prob,SRIW1(),trajectories=100_000);
  3.166502 seconds (69.13 M allocations: 6.577 GiB, 57.88% gc time)
```
After
```julia
julia> @time solve(ensemble_prob,SRIW1(),trajectories=10);
  5.336801 seconds (13.34 M allocations: 977.709 MiB, 6.05% gc time, 99.99% compilation time)

julia> @time solve(ensemble_prob,SRIW1(),trajectories=100_000);
  2.892275 seconds (68.33 M allocations: 6.527 GiB, 55.66% gc time)
```
Maybe this is too ugly, but we should at least do something about the type instability as a matter of principle.

### DIFF
--- a/docs/src/features/ensemble.md
+++ b/docs/src/features/ensemble.md
@@ -393,9 +393,11 @@ Once again, we do this with a `prob_func`, and here we modify the parameters in
 `prob.p`:
 
 ```julia
-function prob_func(prob,i,repeat)
-  x = 0.3rand(2)
-  remake(prob,p=[p[1:2];x])
+prob_func = let p=p
+    (prob,i,repeat) -> begin
+        x = 0.3rand(2)
+        remake(prob, p=[p[1], p[2], x])
+    end
 end
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/features/ensemble.md
```diff
@@ -393,9 +393,11 @@ Once again, we do this with a `prob_func`, and here we modify the parameters in
 `prob.p`:
 
 ```julia
-function prob_func(prob,i,repeat)
-  x = 0.3rand(2)
-  remake(prob,p=[p[1:2];x])
+prob_func = let p=p
```
ChrisRackauckas, line 396: make a comment as to why

PR #568, opened 2022-07-26
Update ode_example.md

troysgit:
Two very small proposed changes: there was a dead hyperlink in the first paragraph whose address was, "https://github.com/JuliaDiffEq/SciMLTutorials.jl", which has been amended to "https://github.com/SciML/SciMLTutorials.jl". 
Another possible change in the same sentence is the addition of a link to the DifferentialEquations.jl site at, "https://diffeq.sciml.ai/stable/" which I think is a great source and layout of information to link new users to but that is up to maintainers to decide.


ChrisRackauckas:
Yes thanks!

PR #578, opened 2022-08-31
add sparse jacobian section to FAQ

taylormcd:
This is a followup to the discussion from https://github.com/SciML/OrdinaryDiffEq.jl/issues/1746


ChrisRackauckas:
Can you include the error message, or part of it? That would help people searching for it.

PR #588, opened 2022-10-08
Document iterator.iter field

gwater:
based on https://discourse.julialang.org/t/callback-at-every-3-integrator-steps-in-differential-equations-jl/56775/5


ChrisRackauckas:
To do this, we need Sundials to get a getproperty overload to make `iter` appear there properly.

gwater:
Ah, so maybe in the meantime something like

>
>`iter` - the number of steps so far (only available for `OrdinaryDiffEq` solvers)
>

ChrisRackauckas:
No, if we commit it to the interface, then it should be made uniform throughout the interface.

PR #590, opened 2022-10-31
document precompilation preferences

ranocha:
I couldn't find a documentation of this before (except in the blog post).


ranocha:
@ChrisRackauckas I also mentioned precompilation of the low-storage methods introduced in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1786. So we should first get a new release of OrdinaryDiffEq.jl before releasing this.

PR #593, opened 2022-11-07
trim down low dep

ArnoStrouwen:


ChrisRackauckas:
I think this is a bit too trim. Some kind of example is definitely required.

ArnoStrouwen:
But the example that I removed is no longer valid.
DiffEqParameterEstim is no longer reexported?

ChrisRackauckas:
We should make a new example then. I think it needs some kind of demo though. I don't think most users will understand it without a demo.

PR #597, opened 2022-11-15
Docs

ArnoStrouwen:
Can't get the `@which` macro to work inside the `@example` block.
https://github.com/SciML/DiffEqDocs.jl/actions/runs/3468601976/jobs/5794598796#step:5:14


ChrisRackauckas:
`@which` requires the standard library InteractiveUtils which is loaded by default in the REPL but not in the doc builds. So you need to add a dependency for InteractiveUtils and explicitly `using InteractiveUtils`.

ArnoStrouwen:
That works but now it looks like:
Capture

PR #642, opened 2023-01-21
Update DAE Problems documentation

owiecc:
Fixes `importdaeproblems not defined` error. `DAEProblemLibrary` does not export this function. 

Made the doc more self-sustained by adding `using Sundials`.


ChrisRackauckas:
> Made the doc more self-sustained by adding using Sundials.

That seems to be missing?

PR #656, opened 2023-03-07
Add example for SecondOrderODEProblem

KronosTheLate:
Some consistensy could problably be worked harder on. But this form also needs documentation.

### DIFF
--- a/docs/src/solvers/dynamical_solve.md
+++ b/docs/src/solvers/dynamical_solve.md
@@ -18,6 +18,11 @@ These correspond to partitioned equations of motion:
 \frac{du}{dt} = f_2(v) \\
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/solvers/dynamical_solve.md
```diff
@@ -18,6 +18,11 @@ These correspond to partitioned equations of motion:
 \frac{du}{dt} = f_2(v) \\
 ```
 
+or, for `SecondOrderODEProblem`,
+```math
+\frac{d^u}{dt^2} = f(d0,u,p,t)
```
ChrisRackauckas, line 23:
```suggestion
\frac{d^2u}{dt^2} = f(u,p,t)
```

PR #677, opened 2023-07-03
Add BVP test problems docs

ErikQQY:
This PR add the BVP test problems docstrings to docs.


ErikQQY:
@ChrisRackauckas bump

PR #683, opened 2023-08-26
Fix plot for EnsembleSummary

ErikQQY:
Instead of using ```plot(summ; idxs=3)```, we should use ```plot(summ; trajectories=3)```

### DIFF
--- a/docs/src/features/ensemble.md
+++ b/docs/src/features/ensemble.md
@@ -589,14 +589,14 @@ the differential equation, this can get messy, so let's only plot the
 
 ```@example ensemble4
 using Plots;
-plot(summ; idxs = 3);
+plot(summ; trajectories = 3);
 ```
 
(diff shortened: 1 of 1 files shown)


rveltz:
I am not sure. I thought it would plot the statistics of the third component of the Problem, not the third trajectory

ErikQQY:
> I thought it would plot the statistics of the third component of the Problem

Yeah, it plots the statistics of the third component of the problem, to plot the "pure" trajectory, we need to set ```error_style``` as ```:none```: ```plot(summ; trajectories=3, error_style=:none)```

ChrisRackauckas:
Yes indeed, this is supposed to be the 3rd index of each solution, not the 3rd trajectory's solution over all variables. It seems that the `idx` indexing in the ensemblesolution is what's having an issue. 

https://github.com/SciML/DifferentialEquations.jl/issues/976#issuecomment-1690185664

PR #702, opened 2023-10-23
update docs to state most derivatives have been implemented

oxinabox:
This note was written 4 years ago.
From talking to @ChrisRackauckas  the majority of them are now implemented.
So we should update the docs so people's default assumption is that things will work.


ChrisRackauckas:
Some of these little notes do need an update 😅

PR #710, opened 2023-12-14
More minor updates

navdeeprana:
Used `!` in functions, slight modification of the text.


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks! that is indeed quite an improvement. I'm not going to worry about the docs build right now since that's unrelated.

PR #711, opened 2023-12-15
Minor improvements

navdeeprana:
@ChrisRackauckas I am just rewriting parts of the documentation as I read along. Mostly rephrasing things I find confusing at first read.

I need your input for the following. Specifying `adaptive = false` in `solve` turns off adapting time step over the course of numerical integration, but it still chooses the initial time step on it's own right?

### DIFF
--- a/docs/src/tutorials/sde_example.md
+++ b/docs/src/tutorials/sde_example.md
@@ -45,12 +45,11 @@ plot(sol)
 ### Using Higher Order Methods
 
 One unique feature of DifferentialEquations.jl is that higher-order methods for
-stochastic differential equations are included. We can compare accuray of 
-the `EM()` method and the higher-order `SRIW1()` method with the analytical solution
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
> I need your input for the following. Specifying adaptive = false in solve turns off adapting time step over the course of numerical integration, but it still chooses the initial time step on it's own right?

I believe if adaptive=false then you need to supply `dt`, though I haven't checked that in a long time. `dt` is the initial time step and if it's not adaptive then it's all time steps.

navdeeprana:
Got it, I missed that we need to supply `dt` with `adaptive=false` [here](https://github.com/SciML/DiffEqDocs.jl/blob/6d14fd899a2795763e2a7178acf8c2efdeaa10bd/docs/src/tutorials/sde_example.md?plain=1#L71C1-L71C1).

PR #719, opened 2024-01-15
Change notation in Brusselator problem setup

navdeeprana:
Changed `u,v` to `U,V` in problem description as `u` is used in diffeq code


ChrisRackauckas:
good point!

PR #729, opened 2024-03-06
Update spiking_neural_systems.md

jbrea:
It looks odd, when the membrane potential in an integrate-and-fire neuron overshoots the threshold.

Btw: I find it convenient to use `ComponentArrays` for parameters, such that I can write e.g. `integrator.p.Vth` instead of `integrator.p[4]`, but I didn't apply this change here, as it would introduce a new dependency.


ChrisRackauckas:
I think it would be fine to introduce the dependency. I would say it's good style and nice to show in an example.

PR #779, opened 2025-05-03
SDE Examples: minor rewording of explanation of "diagonal vs non-diagonal noise"

hurak:
I tried to polish the explanation of the diagonal vs non-diagonal noise in SDE examples. In particular, I found it confusing that the function `g` in the code for the diagonal noise returned/modified a vector, while the equally named g in the equation was a matrix. That the function implicitly returns/modifies a diagonal of the matrix was a fact that could have escaped one's attention. I do confirm that it was stated in the documentation but I just tried to emphasise it even more.

While working on the rewording, I also replaced in the codes the names of in-place functions `f` and `g` with `f!` and `g!`. I hope this will not be found as annoying nitpicking. As I am checking other examples in DifferentialEquations.jl, while some obviously follow this convention, majority does not.

### DIFF
--- a/docs/src/tutorials/sde_example.md
+++ b/docs/src/tutorials/sde_example.md
@@ -268,7 +268,7 @@ A[2, 4] = 1
 A = sparse(A)
 
 # Make `g!` write the sparse matrix values
-function g(du, u, p, t)
+function g!(du, u, p, t)
     du[1, 1] = 0.3u[1]
     du[1, 4] = 0.12u[2]
     du[2, 4] = 1.8u[2]
--- a/docs/src/tutorials/sde_example.md
+++ b/docs/src/tutorials/sde_example.md
@@ -275,7 +275,7 @@ function g(du, u, p, t)
 end
 
 # Make `g!` use the sparse matrix
-prob = SDEProblem(f, g, ones(2), (0.0, 1.0), noise_rate_prototype = A)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

docs/src/tutorials/sde_example.md
```diff
@@ -271,18 +267,18 @@ A[1, 4] = 1
 A[2, 4] = 1
 A = sparse(A)
 
-# Make `g` write the sparse matrix values
+# Make `g!` write the sparse matrix values
 function g(du, u, p, t)
```
ChrisRackauckas, line 271:
```suggestion
function g!(du, u, p, t)
```

docs/src/tutorials/sde_example.md
```diff
@@ -271,18 +267,18 @@ A[1, 4] = 1
...
     du[1, 1] = 0.3u[1]
     du[1, 4] = 0.12u[2]
     du[2, 4] = 1.8u[2]
 end
 
-# Make `g` use the sparse matrix
+# Make `g!` use the sparse matrix
 prob = SDEProblem(f, g, ones(2), (0.0, 1.0), noise_rate_prototype = A)
```
ChrisRackauckas, line 278:
```suggestion
prob = SDEProblem(f!, g!, ones(2), (0.0, 1.0), noise_rate_prototype = A)
```

PR #797, opened 2025-07-23
Update return code documentation to reflect enum implementation

### LINKED ISSUE #734: Return code documentation describes symbols instead of enum
Hi,

The documentation for solution return codes at https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#retcodes describes symbols as return codes, which I believe is out of date and does not match the enum implementation described at https://docs.sciml.ai/SciMLBase/stable/interfaces/Solutions/#retcodes?

ChrisRackauckas:
## Summary
This PR updates the return code (retcode) documentation to reflect the current enum-based implementation instead of the outdated symbol-based approach.

## Changes
- Updated `docs/src/basics/solution.md` to explain that return codes are now enum values
- Added proper usage example showing `SciMLBase.successful_retcode(sol)`
- Added warning about the deprecated `sol.retcode == :Success` pattern
- Fixed incorrect usage in `docs/src/basics/faq.md` (was passing `sol.retcode`, should be just `sol`)
- Added link to the SciMLBase ReturnCode documentation for complete reference

## Context
As pointed out in issue #734, the documentation was outdated and still described return codes as symbols (e.g., `:Success`) when they are now implemented as an enum using EnumX.jl.

## Example
The documentation now shows the correct way to check for successful solutions:
```julia
# Correct (new way)
SciMLBase.successful_retcode(sol)

# Incorrect (old way - now deprecated)
sol.retcode == :Success
```

Fixes #734

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/src/basics/solution.md
+++ b/docs/src/basics/solution.md
@@ -196,7 +196,8 @@ SciMLBase.successful_retcode(sol)
     as successful. For example, `Terminated` is a successful run to a manual termination, and would be missed
     if only checking for Success. Therefore we highly recommend you use `SciMLBase.successful_retcode(sol)` instead.
 
-The return codes include:
+The return codes include are accessed via the ReturnCode module, i.e. `SciMLBase.ReturnCode.Success`. The
+following are major return codes to know:
 
   - `Default`: The solver did not set retcodes.
   - `Success`: The integration completed without erroring or the steady state solver
--- a/docs/src/basics/solution.md
+++ b/docs/src/basics/solution.md
@@ -181,23 +181,38 @@ SciMLBase.DEStats
 
 ## [Return Codes (RetCodes)](@id retcodes)
 
-The solution types have a `retcode` field which returns a symbol signifying the
-error state of the solution. The retcodes are as follows:
+The solution types have a `retcode` field which returns an enum value signifying the
+error state of the solution. Return codes are now implemented as an enum using EnumX.jl
+rather than symbols.
 
-  - `:Default`: The solver did not set retcodes.
-  - `:Success`: The integration completed without erroring or the steady state solver
+To check if a solution was successful, use:
+```julia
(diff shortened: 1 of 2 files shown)


docs/src/basics/solution.md
```diff
@@ -181,23 +181,38 @@ SciMLBase.DEStats
...
-  - `:Default`: The solver did not set retcodes.
-  - `:Success`: The integration completed without erroring or the steady state solver
+To check if a solution was successful, use:
+```julia
+SciMLBase.successful_retcode(sol)
+```
+
+!!! warning
+    Previous iterations of the interface suggested using `sol.retcode == :Success`, 
+    however, that is now not advised. Use `SciMLBase.successful_retcode(sol)` instead.
```
ChrisRackauckas, line 195:
```suggestion
    however, that is now not advised because there are more than one return code that can be interpreted
    as successful. For example, `Terminated` is a successful run to a manual termination, and would be missed
    if only checking for Success. Therefore we highly recommend you use `SciMLBase.successful_retcode(sol)` instead.
```

docs/src/basics/solution.md
```diff
@@ -181,23 +181,40 @@ SciMLBase.DEStats
...
+SciMLBase.successful_retcode(sol)
+```
+
+!!! warning
+    Previous iterations of the interface suggested using `sol.retcode == :Success`, 
+    however, that is now not advised because there are more than one return code that can be interpreted
+    as successful. For example, `Terminated` is a successful run to a manual termination, and would be missed
+    if only checking for Success. Therefore we highly recommend you use `SciMLBase.successful_retcode(sol)` instead.
+
+The return codes include:
```
ChrisRackauckas:
```suggestion
The return codes include are accessed via the ReturnCode module, i.e. `SciMLBase.ReturnCode.Success`. The
following are major return codes to know:
```

PR #850, opened 2026-04-13
Update docs for SciMLBase v3

ChrisRackauckas-Claude:
## Summary

Ports the user-facing documentation to the SciMLBase v3 API, as a companion to the ecosystem-wide v3 compat sweep.

### `docs/src/features/ensemble.md` (main surface)

- Constructor signature: `EnsembleProblem(prob::DEProblem; output_func = (sol, i) -> …, prob_func = (prob, i, repeat) -> …)` → `prob::AbstractDEProblem`, `(sol, ctx)`, `(prob, ctx)`.
- All `prob_func(prob, i, repeat)` function definitions in the `@example` blocks (random ICs, multithreading, pre-determined ICs, SDE parameter study, reduction/convergence example — 6 sites) → `prob_func(prob, ctx)`.
- All `output_func(sol, i)` definitions → `output_func(sol, ctx)`.
- Inline references to `i` / `repeat` inside the example bodies → `ctx.sim_id` / `ctx.repeat`.
- Prose walkthrough of the `prob_func` and `output_func` semantics rewritten to describe the `EnsembleContext` fields (`ctx.sim_id`, `ctx.repeat`, `ctx.rng`, `ctx.sim_seed`, `ctx.master_rng`) instead of the old positional `i` / `repeat`.
- One `last(sol)` usage in the convergence example replaced with `sol.u[end]` — single-int indexing on non-scalar timeseries solutions changed semantics in v3, so `last(sol)` (which falls back to iterating) is no longer reliably the final state vector.

### `docs/src/basics/common_solver_opts.md`
- `solve(prob::SciMLBase.DEProblem,…)` → `solve(prob::SciMLBase.AbstractDEProblem,…)`. The bare `DEProblem` type alias was removed in v3; `AbstractDEProblem` is the supported abstract supertype.

### `docs/src/features/ensemble.md` (constructor signature)
- `EnsembleProblem(prob::DEProblem; …)` → `prob::AbstractDEProblem` (same rationale as above).

### `docs/src/basics/integrator.md`
- Added `derivative_discontinuity!` to the stepping-controls `@docs` block alongside `u_modified!`. SciMLBase v3 renames `u_modified!` → `derivative_discontinuity!` with an `@deprecate` shim, so both docstrings resolve; including both gives users a pointer to the new name without breaking lookups from v2 documentation.

### `docs/src/features/callback_functions.md`
- `DE.u_modified!(integrator, false)` → `DE.derivative_discontinuity!(integrator, false)` in the `AutoAbstolAffect` example. Calls the preferred v3 name; on v2 this symbol doesn't exist, which is consistent with the bumped compat floor.

### `docs/Project.toml`
- `SciMLBase = "2.81.0"` → `"2.81.0, 3.1"`.

## Known limitations

- **Doc build will fail at `Pkg.resolve` until upstream unblocks.** SciMLBase 3.1 requires `RecursiveArrayTools` v4, but `OrdinaryDiffEq` in the registry still pins RAT ≤ 3.54. Every `@example` block that touches ODE/SDE/DDE solve will fail at env instantiation until OrdinaryDiffEq releases a RAT-v4-compatible version. Re-running docs CI after that should go green.
- **Ensemble examples are now v3-only.** v3 removed the 2-arg vs 3-arg arity detection for `prob_func`/`output_func`, so the new syntax in these examples will error if a doc build resolves to SciMLBase v2. The compat entry `"2.81.0, 3.1"` therefore becomes effectively "v3-only at build time" for the ensemble page — consider tightening to `"3.1"` if you'd rather make that explicit.
- **Not tested locally** for the same resolution reason. CI on this PR is the only venue where the rebuilt docs can be exercised.

## Context

Part of a courtesy sweep across the SciMLBase v3 reverse-dependency tree. Related PRs:

- SciML/DiffEqCallbacks.jl#300 — `u_modified!` → `derivative_discontinuity!` via `isdefined` shim (reference pattern for the rename).
- SciML/DiffEqParamEstim.jl#290 — `DEProblem` → `AbstractDEProblem`, ensemble port, non-scalar `sol[i]` → `sol.u[i]`.
- Smaller courtesy PRs: gregoirepourtier/SkeelBerzins.jl#17, hersle/SymBoltz.jl#80, j-fu/VoronoiFVMDiffEq.jl#6.
- Comments flagging the `u_modified!` hot paths on existing compat PRs at trixi-framework/Trixi.jl#2918, #2919; trixi-framework/TrixiParticles.jl#1134; WIAS-PDELib/VoronoiFVM.jl#266; TuringLang/Turing.jl#2802.

cc @ChrisRackauckas

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/src/basics/integrator.md
```diff
@@ -195,6 +195,7 @@ full_cache
 ### [Stepping Controls](@id stepping_controls)
 
 ```@docs
+derivative_discontinuity!
 u_modified!
```
ChrisRackauckas:
```suggestion
```

PR #863, opened 2026-05-28
Re-instantiate Pkg registry in Documentation.yml

abhro:
## Checklist

- [x] Appropriate tests were added
- [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).
  
## Additional context

Maybe that will fix the GitHub workflow's problem on resolving versions when building docs (https://github.com/SciML/DiffEqDocs.jl/actions/runs/26536151206/job/78165442948). First waiting on the CI checks to see if it even works, because on a local machine I was unable to reproduce the error.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -24,12 +24,16 @@ jobs:
       - uses: julia-actions/setup-julia@v3
         with:
           version: "1"
+      - name: Set up Julia registries
+        shell: julia --color=yes --project=docs {0}
+        run: using Pkg; Pkg.Registry.rm("General"); Pkg.Registry.add()
+        env:
+          JULIA_PKG_SERVER: ''
       - name: Set up environment
         shell: julia --color=yes --project=docs {0}
         run: |
           println("--- Instantiating project")
           using Pkg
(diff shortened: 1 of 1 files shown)


abhro:
So that worked, but now I'm concerned that it may be overwriting global registries on the CI machine? (Because Pkg registries on Julia aren't environment specific.) But on the other hand, if the machine is only for CI, then any job would probably just re-add a registry or whatever so we don't need to worry about that.

abhro:
Well, the documentation build fails in a different way now, so this PR succeeded in its aims! :smiling_face_with_three_hearts:

ChrisRackauckas:
I don't think this is the solution. The runner should just get its registry update unblocked.

abhro:
I'm not sure what you mean by the "registry update unblocked"?

ChrisRackauckas:
We have the new runners online, but just need the pool to be opted into here.

ChrisRackauckas:
With the new runner pool we shouldn't need to do this.

PR #903, opened 2026-09-03
Put different classical physics examples in different namespaces

abhro:
## Checklist

- [x] Appropriate tests were added
- [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).

## Additional context

The different examples have different workspace setups anyway, and don't share any commonalities other than being on the same page.


ChrisRackauckas:
Landed as #904, which is this change plus the one import it needed to build: the final henon-heiles block calls `ODERKN.DPRKN6()`, but `OrdinaryDiffEqRKN` was imported only in the harmonic-oscillator block — and that block became anonymous here, so it gets its own module and the name never reaches henon-heiles. That was the `docs/make.jl:175` failure. You are credited as co-author on the merged commit. Thanks!

PR #908, opened 2026-09-12
Document solving the stiff Brusselator on the GPU with a sparse Jacobian

AJ0070:
Adds a GPU section to the advanced ODE tutorial, after the preconditioner section, since it uses the same Brusselator and the same sparse Jacobian setup.

Two things have to change to run that problem on a device: the right-hand side has to broadcast over an index set that lives on the device, because `brusselator_2d_loop` uses a scalar `for` loop, and the `jac_prototype` has to be a `CuSparseMatrixCSR` so the Jacobian is built on the device too. Preconditioners work through `precs` with `kp_ilu0` from KrylovPreconditioners.jl.

Checked on a T4 at the tutorial's own `(32, 32, 2)` state and sparsity prototype: TRBDF2 and FBDF both match the CPU reference, unpreconditioned GMRES to 8e-10 and `kp_ilu0` through `precs` to 6e-8. The blocks are not executed by the docs build, matching the Sundials preconditioner examples on the same page, since CUDA is not a docs dependency.

Requested in SciML/OrdinaryDiffEq.jl#1566, replacing SciML/OrdinaryDiffEq.jl#4453. Needs SciML/OrdinaryDiffEq.jl#4452, which is merged.

AI Disclosure: Used Opus 5


docs/src/tutorials/advanced_ode_example.md
```diff
@@ -331,6 +331,104 @@ nothing # hide
...
 
+## Solving on the GPU
+
+The sparse Jacobian setup above works on a GPU, but two things have to change.
+
+The first is the right-hand side. `brusselator_2d_loop` walks the grid with a scalar
+`for` loop, and that cannot run on a device. Writing the same stencil as one broadcast
+puts the work on whichever device the index set lives on:
+
+```julia
```
ChrisRackauckas: This doesn't build / run the example.

